PR #86, opened 2019-07-09
fixed property name and initialized data variable in readme

mhstoller:
corrected video duration property name to videoDuration


CaiJingLong:
LGTM

Issue #60, opened 2019-03-10
[BUG] IOS 12.1.4 Crash problem


zzpmaster:
**Describe the bug**
打开相册，ios崩溃

导致崩溃代码
```
ImageScanner.m

1. for (PHCollection *collection in self.galleryArray) {
2.     PHFetchResult<PHAssetCollection *> *fetchResult = [PHAssetCollection
3.             fetchAssetCollectionsWithLocalIdentifiers:@[collection.localIdentifier] options:opt];
4.     for (PHAssetCollection *assetCollection in fetchResult) {
5. 
6.         PHFetchResult<PHAsset *> *assetResult = [PHAsset
7.                 fetchAssetsInAssetCollection:assetCollection options:opt];
8.         for (PHAsset *asset in assetResult) {
9.                 _idAssetDict[asset.localIdentifier] = asset;
10.              block(assetCollection, asset);
11.         }
12.     }
13. }

```
assetCollection 有可能是PHCollectionList对象，需要过滤掉, 需要在问题代码**第五行**添加下面代码。
```
if (![assetCollection isKindOfClass:[PHAssetCollection class]]) continue;
```
修改后可以正常运行

**请验证，如果正确期待回复和更新。**

**Flutter version**
[✓] Flutter (Channel beta, v1.2.1, on Mac OS X 10.14.3 18D109, locale en-US)
[✓] Android toolchain - develop for Android devices (Android SDK version 28.0.3)
[✓] iOS toolchain - develop for iOS devices (Xcode 10.1)
[✓] Android Studio (version 3.2)
[✓] VS Code (version 1.31.1)
[✓] Connected device (1 available)

**Smartphone (please complete the following information):**
 - Device: iPhone X
 - OS: 12.1.4
 - Version 0.3.2

**Log**

`2019-03-10 21:30:30.916729+0800 Runner[12455:3105536] -[PHCollectionList assetCollectionType]: unrecognized selector sent to instance 0x281a99080

2019-03-10 21:30:37.159769+0800 Runner[12455:3105536] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[PHCollectionList assetCollectionType]: unrecognized selector sent to instance 0x281a99080'
*** First throw call stack:
[...]

PR #93, opened 2019-08-28
another fix for crash of same cause as #60

rcunning:
I got the same crash as #60 but in a different call. It is unclear to me how to reproduce the issue as it has something to do with what albums the phone has on it. The phone that was crashing has many many albums, one of which must have been causing the issue.


CaiJingLong:
This can only guarantee that it will not crash, and it will not be able to get a deeper level of image. There will be some logical problems.
When I have time, I will fix this problem on the basis of your PR.

rcunning:
Thank you!

Issue #91, opened 2019-08-19
[Feature] Paginated asset path list


SvenSchoene:
**Is your feature request related to a problem? Please describe.**
- On an iPhone 6 with 8.500+ assets it takes roughly  25 seconds for `getAssetPathList()` to load.
- On an iphone 8 with 25.000+ assets the app crashes with this exception taken from Firebase Crashlytics:

```
Fatal Exception: NSRangeException
*** -[__NSArray0 objectAtIndex:]: index 0 beyond bounds for empty NSArray

Fatal Exception: NSRangeException
0  CoreFoundation                 0x1cf0e127c __exceptionPreprocess
1  libobjc.A.dylib                0x1ce2bb9f8 objc_exception_throw
2  CoreFoundation                 0x1ceff67bc -[__NSCFArray count]
3  Photos                         0x1ddb2500c -[PHFetchResult objectAtIndex:]
4  photo_manager                  0x1032f4b34 (Missing)
5  photo_manager                  0x1032f47a4 (Missing)
6  libdispatch.dylib              0x1ceb20a38 _dispatch_call_block_and_release
7  libdispatch.dylib              0x1ceb217d4 _dispatch_client_callout
8  libdispatch.dylib              0x1ceafcdec _dispatch_lane_serial_drain$VARIANT$armv81
9  libdispatch.dylib              0x1ceafd92c _dispatch_lane_invoke$VARIANT$armv81
10 libdispatch.dylib              0x1ceb05e08 _dispatch_workloop_worker_thread
11 libsystem_pthread.dylib        0x1ced01114 _pthread_wqthread
12 libsystem_pthread.dylib        0x1ced03cd4 start_wqthread
```

and
```
Fatal Exception: NSRangeException
*** -[__NSArray0 objectAtIndex:]: index 0 beyond bounds for empty NSArray

Fatal Exception: NSRangeException
0  CoreFoundation                 0x1cf0e127c __exceptionPreprocess
1  libobjc.A.dylib                0x1ce2bb9f8 objc_exception_throw
2  CoreFoundation                 0x1ceff67bc -[__NSCFArray count]
3  Photos                         0x1ddb2500c -[PHFetchResult objectAtIndex:]
4  photo_manager                  0x1032f4b34 (Missing)
5  photo_manager                  0x1032f47a4 (Missing)
6  libdispatch.dylib              0x1ceb20a38 _dispatch_call_block_and_release
[...]

PR #94, opened 2019-08-30
Feature/paginated asset path list

2ZeroSix:
Partly implemented feature mentioned in this issue #91

Original issue proposed paginated version of `getAssetPathList`. But I noticed, that real fetch is happening in `_getAssetList`. Therefore, I implemented a `_getAssetListPaged`.

It's only works for AssetPathEntity with (onlyImage || onlyVideo == false). This behavior is caused by the current implementation of getVideoPathList and getImagePathList: on ios it fetches and filters out collections if they doesn't contain assets of corresponding type. I didn't find a way to filter them fast, so pagination wouldn't solve the performance problem for this type of assets.

On android side I changed `scanAndGetImageIdList` implementation a bit:
- `scan()` => `scanBuckets()`
- and added deferred call to `scan()` in `getImageListWithPathId` and `getAllImageList`

Then i changed `getImageWithId` implementation, now it fetches Asset if it's not in the cache.

These changes were made at the request of the creator of the #91 issue and are used by him in his project.

If something is unclear or does not match your vision of this feature, I'm ready to discuss.

### DIFF
--- a/android/src/main/kotlin/top/kikt/imagescanner/ImageScanner.kt
+++ b/android/src/main/kotlin/top/kikt/imagescanner/ImageScanner.kt
@@ -97,7 +97,7 @@ class ImageScanner(private val registrar: PluginRegistry.Registrar) {
         videoCursor.close()
     }
 
-    private fun handleBucketCursor(mCursor: Cursor, isImage: Boolean = false, isVideo: Boolean=false) {
+    private fun handleBucketCursor(mCursor: Cursor, isImage: Boolean = false, isVideo: Boolean = false) {
         val dir = mCursor.getString(mCursor.getColumnIndex(MediaStore.Images.ImageColumns.BUCKET_DISPLAY_NAME))
         val dirId = mCursor.getString(mCursor.getColumnIndex(MediaStore.Images.ImageColumns.BUCKET_ID))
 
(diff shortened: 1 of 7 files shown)


CaiJingLong:
Sorry, I haven’t responded for so long, thank you for your PR.

I will test this PR as soon as possible.

SvenSchoene:
FWIW: We're currently using this on all our tests devices internally, and it works very, very well.

I understand if this is not how you would have intended to implement this yourself, maybe. But I just wanted to give some feedback, that this (1) works reliably, (2) has a great memory profile and (3) is still very fast, both, (3.1) loading a page of files, as well as (3.2) generating thumbnails.

CaiJingLong:
Some descriptions may be needed to help the user quickly understand how to use it.
For example, how to use the page attribute (starting at 0 or 1).

The description can be a method comment or a readme document.

If you don't have time to add, I can complete this part.

CaiJingLong:
The code section looks very good, I will merge this part of the code after 2 days, if you are convenient, please help to add the document, if you do not have time to add the document, I will complete this part after the merger.

SvenSchoene:
I won't be able to contribute any commits, but some info:

### Page numbering
Pages start at `0`. I was wondering why I was missing the newest assets, but I was always starting at `1`. So you're right: this is an important note to add.

### How to use

Here is one, simplified version of what we're using:

We're using the **flutter_pagewise** library for displaying a grid in this example: https://pub.dev/packages/flutter_pagewise

```dart
AssetPathEntity allAssetsDirectory;

void initAllAssetsDirectory() async {
  List<AssetPathEntity> directories = await PhotoManager.getAssetPathList();
  if (directories.length < 1) return null;

  // the first directory is a special directory and contains all medias.
  allAssetsDirectory = directories[0];
}

Future<List<AssetEntity>> loadMore(int page) async {
  return await allAssetsDirectory.getAssetListPaged(page, perPage);
}

@override
Widget build(BuildContext context) {
  return PagewiseGridView.count(
    key: gridKey,
    pageSize: 40,
    crossAxisCount: 3,
    mainAxisSpacing: 1.0,
    crossAxisSpacing: 1.0,
    shrinkWrap: true,
    padding: EdgeInsets.all(0),
    itemBuilder: (BuildContext context, AssetEntity a, int index) {
      ValueKey key = ValueKey(m.path);

      return MyOwnGridTileWidget(
        key: key,
        asset: a,
      );
    },
    pageFuture: (int page) {
      return loadMore(page);
    },
  );
}
```

CaiJingLong:
OK, I will merge this part of the code to the master as soon as possible, and I will add the corresponding document.

Thanks for PR.

CaiJingLong:
Since image/video is not supported, I will further modify it recently to support flag.

PR #114, opened 2019-09-22
Fix null duration

rignaneseleo:
It avoids the error: The method '_mulFromInteger' was called on null.

Here I explain my case: https://github.com/CaiJingLong/flutter_photo/issues/66


rignaneseleo:
Is it already available from pubspec?

CaiJingLong:
Version 0.4.3 contains this PR.

PR #152, opened 2020-01-13
Add location info of AssetEntity.

CaiJingLong:


CaiJingLong:
During the writing process, fixed a save asset bug on androidQ devices,

CaiJingLong:
Because location needs to get the original file in androidQ,
Therefore, the method of obtaining the original file in iOS has also been completed.

PR #213, opened 2020-03-19
Add orientation.

CaiJingLong:


mhstoller:
To clarify this change, the library will not report the aspect ratio using the orientation automatically right? In my code I must check the AssetEntity's orientation property and swap the height and width if the orientation is 90 or 270?

I just want to make sure if I implement these checks in my code it will not change in the future and cause me other issues :)

PR #230, opened 2020-03-24
Ignore size

CaiJingLong:


CaiJingLong:
Even if the width height is empty, you can scan to the asset by setting the `sizeConstraint.ignoreSize` property of `imageOption`.

PR #235, opened 2020-03-27
Completed some fields for Android's saveVideo.

CaiJingLong:


CaiJingLong:
As duration width size title display_name

Issue #315, opened 2020-06-20
[Feature] Add ability to set deliveryMode


jlavyan:
**Is your feature request related to a problem? Please describe.**

On get thumb image we set size, but iOS give bigger image, for example if we put 400x400 we will get > 400x400 image

**Describe the solution you'd like**
    PHImageRequestOptions has deliveryMode which consider below enums

    @available(iOS 8, *)
    case opportunistic = 0 // client may get several image results when the call is asynchronous or will get one result when the call is synchronous

    @available(iOS 8, *)
    case highQualityFormat = 1 // client will get one result only and it will be as asked or better than asked

    @available(iOS 8, *)
    case fastFormat = 2 // client will get one result only and it may be degraded

By default we use PHImageRequestOptionsDeliveryModeOpportunistic, for my case i need fastFormat

**Describe alternatives you've considered**
NO

**Additional context**

On scroll grid with images app freezes

jlavyan:
I fixed this issue by set thumbnail contentMode to ResizeContentMode.fit

PR #318, opened 2020-06-24
[PR] ADD image load options

jlavyan:
### ✨ What kind of change does this PR introduce? (Bug fix, feature, docs update...)
Feature

### 💥 Does this PR introduce a breaking change?
No i don't think, i just converted current image load parameters into option

### 📝 Links to relevant issues/docs
#315 

### 🆕 What is the new behavior (if this is a feature change)?
I created load option for make specific parameters in iOS, by default for all case we will use  DefaultLoadOption

### DIFF
--- a/example/lib/page/copy_to_another_gallery_example.dart
+++ b/example/lib/page/copy_to_another_gallery_example.dart
@@ -34,7 +34,8 @@ class _CopyToAnotherGalleryPageState extends State<CopyToAnotherGalleryPage> {
           AspectRatio(
             aspectRatio: 1,
             child: FutureBuilder<Uint8List>(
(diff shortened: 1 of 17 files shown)


CaiJingLong:
Thank you for your work.

My understanding is different from yours. Any API incompatibility for users can be understood as breaking change.

So, when I merge this PR into master, it means that it will be part of version 0.6.

CaiJingLong:
I create new branch 0.6.x, I will develop the next version in this branch.

Can you open Pull Request for the branch 0.6.x?

jlavyan:
Yes sure. This is pull request 

[322](https://github.com/CaiJingLong/flutter_photo_manager/pulls)

PR #322, opened 2020-07-06
[PR] Load image options

jlavyan:
**New thumb image load options for iOS**

### Related  pull
#318

### DIFF
--- a/example/lib/page/copy_to_another_gallery_example.dart
+++ b/example/lib/page/copy_to_another_gallery_example.dart
@@ -34,7 +34,8 @@ class _CopyToAnotherGalleryPageState extends State<CopyToAnotherGalleryPage> {
           AspectRatio(
             aspectRatio: 1,
             child: FutureBuilder<Uint8List>(
-              future: widget.assetEntity.thumbDataWithSize(500, 500),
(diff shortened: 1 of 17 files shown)


CaiJingLong:
I tried to run the project, and when I used the DefaultOption, it seemed to break the previous default behavior.

For a 3000x2002 picture:
1. In the 0.6.x branch, before you commit the code, `entity.thumbDataWithSize (500, 500)` the width and height of the image will get 748x500 thumb.
2. Then, I checkout to the PR branch, and use the `DefaultLoadOption(500,500)`, then the entity.thumbDataWithSize(option) will get 3000x2002 thumb.
3. In your branch, I use the iOSLoadOption(500,500), the method will get 748x500.


I think this is a problem. 

Expect: `DefaultOption (500,500)` should return no less than 500x500 images (such as 748x500).

image

By the way: The test picture I used is one of the pictures that comes with the iOS simulator.
image

jlavyan:
I changed default values in plugin side. So now it will use default parameters as before when deliveryMode, contentMode, resizeMode was not set @CaiJingLong

CaiJingLong:
Good job, thanks for your work.

PR #337, opened 2020-07-20
Open stream in dart

CaiJingLong:
You can "manually" open a Stream in dart to read stream information directly, without creating a local file.
A typical usage method is http file upload.


CaiJingLong:
The code part has been completed, and there may be some use cases that require additional testing.
The document is incomplete.

Issue #287, opened 2020-05-03
Add method to check if storage permission is granted


TheLastGimbus:
We already have `PhotoManager.requestPermission()` for getting permission.

It would be nice to have method to check if permission is granted.

TheLastGimbus:
I wrote this wile writing some code that was using PhotoManager in the background (using [workmanager](https://pub.dev/packages/workmanager)

It keeps giving me exception:

```java
E/MethodChannel#top.kikt/photo_manager( 5686): Failed to handle method call
E/MethodChannel#top.kikt/photo_manager( 5686): java.lang.NullPointerException: 获取权限的Activity不存在
E/MethodChannel#top.kikt/photo_manager( 5686): 	at top.kikt.imagescanner.permission.PermissionsUtils.getPermissionsWithTips(PermissionsUtils.java:110)
E/MethodChannel#top.kikt/photo_manager( 5686): 	at top.kikt.imagescanner.permission.PermissionsUtils.getPermissions(PermissionsUtils.java:95)
E/MethodChannel#top.kikt/photo_manager( 5686): 	at top.kikt.imagescanner.core.PhotoManagerPlugin.onMethodCall(PhotoManagerPlugin.kt:152)
E/MethodChannel#top.kikt/photo_manager( 5686): 	at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler.onMessage(MethodChannel.java:231)
E/MethodChannel#top.kikt/photo_manager( 5686): 	at io.flutter.embedding.engine.dart.DartMessenger.handleMessageFromDart(DartMessenger.java:93)
E/MethodChannel#top.kikt/photo_manager( 5686): 	at io.flutter.embedding.engine.FlutterJNI.handlePlatformMessage(FlutterJNI.java:642)
E/MethodChannel#top.kikt/photo_manager( 5686): 	at android.os.MessageQueue.nativePollOnce(Native Method)
E/MethodChannel#top.kikt/photo_manager( 5686): 	at android.os.MessageQueue.next(MessageQueue.java:326)
E/MethodChannel#top.kikt/photo_manager( 5686): 	at android.os.Looper.loop(Looper.java:160)
E/MethodChannel#top.kikt/photo_manager( 5686): 	at android.app.ActivityThread.main(ActivityThread.java:6718)
E/MethodChannel#top.kikt/photo_manager( 5686): 	at java.lang.reflect.Method.invoke(Native Method)
[...]

PR #361, opened 2020-09-02
Make use of permission_handler instead of our own implementation

TheLastGimbus:
This PR adds [permission_handler](https://pub.dev/packages/permission_handler) as a dependency, and makes use of it. Thanks to this, we will have easier job of messing with permissions, and we can check permissions without Activity (see discussion in #287 )

Changes made by me still need to be tested on iOS, and on new Android 11 !

If @CaiJingLong will decide to accept this, it will also be nice to remove old attempts in Android and iOS sides.

### DIFF
--- a/lib/photo_manager.dart
+++ b/lib/photo_manager.dart
@@ -6,6 +6,7 @@ import 'dart:ui';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/rendering.dart';
 import 'package:flutter/services.dart';
+import 'package:permission_handler/permission_handler.dart';
 import 'package:photo_manager/src/utils/convert_utils.dart';
 import 'src/filter/filter_options.dart';
 
--- a/lib/src/manager.dart
+++ b/lib/src/manager.dart
@@ -10,7 +10,14 @@ class PhotoManager {
   ///
   /// in ios request the photo permission
   static Future<bool> requestPermission() async {
-    return _plugin.requestPermission();
+    if(Platform.isAndroid){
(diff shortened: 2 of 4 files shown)


CaiJingLong:
This PR will not be merged, and referencing other flutter libraries will cause version conflicts.

TheLastGimbus:
Not necessary

When you write `library: ^2.0.0`, pub will accept anything `2.*.*`

So user can have `^2.4.3`, other library can have `^2.0.1`, and everything will work

CaiJingLong:
No, for example, if the plugin you posted is updated to 3.0, plugin 1 depends on 3.0, and plugin 2 depends on 2.0.
When they are used by the same person, they conflict.

----

The library part of this plugin will not depend on other plugins, nor will it in the future

TheLastGimbus:
So what can we do? I really need the library to work in the background - can we just check how `permission_handler` implemented it, and copy-paste the same in here?

CaiJingLong:
I think what you need may be to ignore the permission check, the default must have permissions, and the contentResolver must be available.

CaiJingLong:
I can add an option to support this, but correspondingly, the risk needs to be borne by the user of plugin.

TheLastGimbus:
I can implement everything what is needed to run this plugin in the background - just tell me what needs to be done

CaiJingLong:
I push a change to ignore permission check.
You can try to use 0.6.0-dev.3 version.
And, you can try to use permission_handler to handle permission.

Issue #385, opened 2020-10-19
[Feature] Support iOS 14 limited access permission in permission check


kuhnroyal:
**Is your feature request related to a problem? Please describe.**

I want to display a small banner to the user if the granted permission on iOS 14 is limited to some photos/albums.
Currently it is only possible to determine if it is granted or not.

**Describe the solution you'd like**

Something like this would allow this and is backwards compatible.
```dart
enum PhotoPermission { none, full, limited }

Future<bool> requestPermission() async {
    final permission = await requestPermissionExtended();
    return permission == PhotoPermission.full || permission = PhotoPermission.limited;
}

Future<PhotoPermission> requestPermissionExtended() async {
    final permission = await _channel.invokeMethod("requestPermission");
    return PhotoPermission.values[permission];
}
```

Of course the iOS plugin side needs to be changed to fit this.

Artem2908:
> **Is your feature request related to a problem? Please describe.**
> 
> I want to display a small banner to the user if the granted permission on iOS 14 is limited to some photos/albums.
> Currently it is only possible to determine if it is granted or not.
> 
> **Describe the solution you'd like**
> 
> Something like this would allow this and is backwards compatible.
> 
> ```dart
> enum PhotoPermission { none, full, limited }
> 
> Future<bool> requestPermission() async {
>     final permission = await requestPermissionExtended();
>     return permission == PhotoPermission.full || permission = PhotoPermission.limited;
> }
> 
> Future<PhotoPermission> requestPermissionExtended() async {
>     final permission = await _channel.invokeMethod("requestPermission");
>     return PhotoPermission.values[permission];
> }
> ```
> 
> Of course the iOS plugin side needs to be changed to fit this.

Could you tell me in more detail how to properly use your version?

lyio:
@Artem2908 the proposed solution would provide a additional way of retrieving permission information.

Currently, the method `requestPermission` in the plugin only tells you, whether the user granted permission to the gallery or not.

In iOS14, Apple added a third possibility. Now the user can only grant permissions for a number of hand selected photos. The proposed `requestPermissionExtended` method would be able to tell you that, so that you can display a warning or what have you in your App, reminding the user of why they do not see all their pictures in the gallery.

Usage would be as simple as 

```dart
  final permission = await  PhotoManager.requestPermissionExtended();
  switch(permission) {
    case PhotoPermission.limited:
   case PhotoPermission.full:
    // permission is granted, do whatever you like
[...]

PR #391, opened 2020-11-04
#385-ios14-limited-permission-support

lyio:
+ Add an enum to represent the different permission states
+ Add a new method to request permissions that returns an instance
  of this enum instead of only `true` or `false`
+ Change MethodChannel implementation in the iOS plugin to
  1. Check for iOS 14
  2. Use new `requestAuthorizationForAccessLevel` API
  3. Differentiate between full access and limited access if applicable.
  The implementation for pre-iOS14 stays unchanged.

### DIFF
--- a/ios/Classes/core/PMPlugin.m
+++ b/ios/Classes/core/PMPlugin.m
@@ -38,15 +38,29 @@
   PMManager *manager = self.manager;
 
   if ([call.method isEqualToString:@"requestPermission"]) {
-    [PHPhotoLibrary requestAuthorization:^(PHAuthorizationStatus status) {
-        BOOL auth = PHAuthorizationStatusAuthorized == status;
-        [manager setAuth:auth];
-        if (auth) {
-          [handler reply:@1];
-        } else {
-          [handler reply:@0];
-        }
-    }];
+      if (@available(iOS 14, *)) {
+          [PHPhotoLibrary requestAuthorizationForAccessLevel:PHAccessLevelReadWrite handler: ^(PHAuthorizationStatus status) {
+            id result = 0;
+            BOOL auth = PHAuthorizationStatusAuthorized == status || PHAuthorizationStatusLimited == status;
+            [manager setAuth:auth];
+            if (PHAuthorizationStatusAuthorized == status) {
+                result = @1;
+            }
+            else if (PHAuthorizationStatusLimited == status) {
+                result = @2;
+            }
+            else {
+                result = @0;
+            }
+          [handler reply:result];
+          }];
+    } else {
+        [PHPhotoLibrary requestAuthorization: ^(PHAuthorizationStatus status) {
+            BOOL auth = PHAuthorizationStatusAuthorized == status;
+            [manager setAuth:auth];
+            [handler reply:auth ? @1 : @0];
+        }];
+    }
   } else if ([call.method isEqualToString:@"clearFileCache"]) {
     [manager clearFileCache];
     [handler reply:@1];
--- a/lib/src/manager.dart
+++ b/lib/src/manager.dart
@@ -13,6 +13,10 @@ class PhotoManager {
     return _plugin.requestPermission();
   }
 
+  /// Provide more information about the actual permission being granted.
+  /// This can account for the .limited permission in iOS 14.
+  static Future<PhotoPermission> requestPermissionExtended() async => _plugin.requestPermissionExtended();
+
   static Editor editor = Editor();
 
   /// get gallery list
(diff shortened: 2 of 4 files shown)


CaiJingLong:
And, please use `flutter format lib` to format your dart code.

ios/Classes/core/PMPlugin.m
```diff
@@ -38,15 +38,29 @@ - (void)onMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result {
...
-    [PHPhotoLibrary requestAuthorization:^(PHAuthorizationStatus status) {
-        BOOL auth = PHAuthorizationStatusAuthorized == status;
-        [manager setAuth:auth];
-        if (auth) {
-          [handler reply:@1];
-        } else {
-          [handler reply:@0];
-        }
-    }];
+      if (@available(iOS 14, *)) {
```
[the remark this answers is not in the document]
CaiJingLong: Just like next:
```m
#ifdef  __IPHONE_14_0
// use new api
#else
// use old api
#endif
```

ios/Classes/core/PMPlugin.m
```diff
@@ -38,15 +38,29 @@ - (void)onMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result {
...
+            }
+            else if (PHAuthorizationStatusLimited == status) {
+                result = @2;
+            }
+            else {
+                result = @0;
+            }
+          [handler reply:result];
+          }];
+    } else {
```
CaiJingLong, line 57:
```suggestion
#else
```

ios/Classes/core/PMPlugin.m
```diff
@@ -38,15 +38,29 @@ - (void)onMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result {
...
+        [PHPhotoLibrary requestAuthorization: ^(PHAuthorizationStatus status) {
+            BOOL auth = PHAuthorizationStatusAuthorized == status;
+            [manager setAuth:auth];
+            [handler reply:auth ? @1 : @0];
+        }];
+    }
```
CaiJingLong, line 63:
```suggestion
#endif
```

ios/Classes/core/PMPlugin.m
CaiJingLong, line 41:
```suggestion
#ifdef  __IPHONE_14_0
```

CaiJingLong (changes requested):

ios/Classes/core/PMPlugin.m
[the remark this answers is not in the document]
lyio: @CaiJingLong, if compatibility with Xcode 11 is important, I will change the check as suggested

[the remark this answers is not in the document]
CaiJingLong: Although developers will definitely migrate to the new IDE version in the future, the old IDE cannot be given up yet.
Yes, I think this is more important at the moment.

You can try to use the macro to see if there is a problem in the XCode12+ios14 environment.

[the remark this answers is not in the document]
lyio: @CaiJingLong the answer here https://stackoverflow.com/questions/3339722/how-to-check-ios-version/3339787#3339787 seems to suggest that `@available` was introduced in Xcode 9, which is is pretty old now.

Thinking about this further, I also realized that `#ifdef` is a precompiler directive, which is only evaluated at compile time. Correct me if I'm wrong, but doesn't that mean that if my compile target is ios 14+ the compile strips out everything between the `#else` and `#endif`? And Apple requires you to compile against the latest target, doesn't it? So you would compile against ios 14, but your App can run on ios 13 and below. So the check would be wrong, wouldn't it? Because it is not a runtime check anymore, but depends on your compilation target. So the code would fail on ios 13, wouldn't it?

But maybe I'm just not understanding the precompiler directives correctly.

lyio: Wouldn't this exclude iPad builds?

CaiJingLong: Macro is used in precompiling. This is just to determine whether a macro is defined in your SDK, not if at runtime.
Therefore, when you use xcode14, your IOS SDK contains this macro, but 13.0 does not. It does not conflict with your target. You can try to run it under xcode14's iPad target.

yanivshaked:
@CaiJingLong This is an important feature - deal with iOS 14 limited access. Any plans to complete it?

CaiJingLong:
> @CaiJingLong This is an important feature - deal with iOS 14 limited access. Any plans to complete it?

I update some code in #495.
Once the test is finished, I will push the new version.

lyio:
With #495 merged, this PR is basically obsolete. Especially since it is based on an old version of the library without the nullsafety features and would need some overhaul to work with the changes.

I will close this PR for now.

PR #397, opened 2020-11-16
[iOS] Use edited photo

krokyze:
Related to https://github.com/CaiJingLong/flutter_photo_manager/issues/184#issuecomment-724698564
Tested and now works as expected 👍

PR #406, opened 2020-12-01
Fix edited data

CaiJingLong:
Related #397.

### DIFF
--- a/ios/Classes/core/PHAsset+PHAsset_getTitle.m
+++ b/ios/Classes/core/PHAsset+PHAsset_getTitle.m
@@ -81,6 +81,9 @@
 - (PHAssetResource *)getAdjustResource {
   NSArray<PHAssetResource *> *resources =
       [PHAssetResource assetResourcesForAsset:self];
+  if (resources.count == 0) {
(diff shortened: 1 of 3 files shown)


ios/Classes/core/PHAsset+PHAsset_getTitle.m
```diff
@@ -38,4 +38,114 @@ - (NSString *)title {
...
+    if (self.mediaType == PHAssetMediaTypeImage ||
+        res.type == PHAssetResourceTypeFullSizePhoto) {
+      return res;
+    }
+
+    if (self.mediaType == PHAssetMediaTypeVideo ||
+        res.type == PHAssetResourceTypeFullSizeVideo) {
+      return res;
+    }
+  }
```
krokyze, line 120: Hey as this is adjusted image shouldn't it be `[resources reverseObjectEnumerator]`?

krokyze: @CaiJingLong yep just checked the `0.6.0-dev.7` version and the issue is still persistent.

PR #415, opened 2020-12-18
[iOS] Use edited photo

krokyze:
As I replied under #406 I tested the newest version and it still doesn't return the edited photo.

In this PR:
- Added getAdjustResource for videos too
- Fixed getAdjustResource in case the asset is adjusted to return latest resource


krokyze:
Sorry @CaiJingLong I finally understood my problem. I was using `originBytes`.

PR #459, opened 2021-03-24
Add mime type

CaiJingLong:


CaiJingLong:
The property is only contains in MediaStore of android.

lib/src/utils/convert_utils.dart
```diff
@@ -47,6 +47,10 @@ class ConvertUtils {
         ..latitude = item['lat']
         ..longitude = item['lng'];
 
+      if (item.containsKey('mimeType')) {
```
AlexV525, line 50: Is a direct assignment invalid? `..mimeType = item['mimeType']`

lib/src/utils/convert_utils.dart
```diff
@@ -74,5 +78,11 @@ class ConvertUtils {
     )
       ..latitude = data['lat']
       ..longitude = data['lng'];
+
+    if (data.containsKey('mimeType')) {
```
AlexV525, line 82: Same as above.

Issue #428, opened 2021-01-14
[Question] How to create an app folder when saving a content?


themadmrj:
I need to create an app directory (or album) when saving a content. Right now, when using `PhotoManager.editor.saveImageWithPath(filePath)`, everything is saved to "Pictures" like so:

![Capture](https://user-images.githubusercontent.com/11475495/104636091-1b250100-56a3-11eb-9829-68a84b560752.PNG)

How do I make the manager save the contents to an app folder? I'm basically looking for the android `RELATIVE_PATH` MediaStore column. Is this related to #231?

The file is originally saved like so:

`/storage/emulated/0/Android/data/com.myapp.android/files/DCIM/app_name/CONTENT_1610651148623.webp`

This path is fine for android and is displayed in a separate directory if detected by the `MediaStore`. Is there any way to simply scan a file to update the `MediaStore` with the current path?

PR #462, opened 2021-03-25
Allow specifying a RELATIVE_PATH when saving a file to the MediaStore…

themadmrj:
On Android, currently saving a file to a custom folder is only possible on Android 28 or below. This PR allows saving to a custom folder on Android 29 or above aswell by using the RELATIVE_PATH field of the MediaStore. Example:
`await PhotoManager.editor.saveImageWithPath(file.path, relativePath: "DCIM/MyApp");`

This solves #428.

### DIFF
--- a/android/src/main/kotlin/top/kikt/imagescanner/core/utils/Android30DbUtils.kt
+++ b/android/src/main/kotlin/top/kikt/imagescanner/core/utils/Android30DbUtils.kt
@@ -382,7 +382,7 @@ object Android30DbUtils : IDBUtils {
       typeFromStream = "image/${File(title).extension}"
     } else {
       typeFromStream = URLConnection.guessContentTypeFromStream(inputStream)
-              ?: "image/${File(title).extension}"
+              ?: "image/*"
     }
 
     val uri = MediaStore.Images.Media.EXTERNAL_CONTENT_URI
--- a/android/src/main/kotlin/top/kikt/imagescanner/core/utils/Android30DbUtils.kt
+++ b/android/src/main/kotlin/top/kikt/imagescanner/core/utils/Android30DbUtils.kt
@@ -362,7 +362,7 @@ object Android30DbUtils : IDBUtils {
     androidQCache.saveAssetCache(context, asset, byteArray, true)
   }
 
-  override fun saveImage(context: Context, image: ByteArray, title: String, desc: String): AssetEntity? {
+  override fun saveImage(context: Context, image: ByteArray, title: String, desc: String, relativePath: String): AssetEntity? {
     val (width, height) =
         try {
           val bmp = BitmapFactory.decodeByteArray(image, 0, image.count())
@@ -393,6 +393,7 @@ object Android30DbUtils : IDBUtils {
       put(MediaStore.Images.ImageColumns.WIDTH, width)
       put(MediaStore.Images.ImageColumns.HEIGHT, height)
     }
+    if (relativePath != null) values.put(MediaStore.Images.ImageColumns.RELATIVE_PATH, relativePath)
 
     val contentUri = cr.insert(uri, values) ?: return null
     val outputStream = cr.openOutputStream(contentUri)
@@ -409,7 +410,7 @@ object Android30DbUtils : IDBUtils {
     return getAssetEntity(context, id.toString())
   }
 
-  override fun saveImage(context: Context, path: String, title: String, desc: String): AssetEntity? {
+  override fun saveImage(context: Context, path: String, title: String, desc: String, relativePath: String): AssetEntity? {
     val cr = context.contentResolver
     val timestamp = System.currentTimeMillis() / 1000
     val inputStream = FileInputStream(path)
@@ -441,6 +442,7 @@ object Android30DbUtils : IDBUtils {
       put(MediaStore.Images.ImageColumns.WIDTH, width)
       put(MediaStore.Images.ImageColumns.HEIGHT, height)
     }
+    if (relativePath != null) values.put(MediaStore.Images.ImageColumns.RELATIVE_PATH, relativePath)
 
     val contentUri = cr.insert(uri, values) ?: return null
     val outputStream = cr.openOutputStream(contentUri)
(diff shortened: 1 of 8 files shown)


AlexV525:
Thanks for the PR. Could you provide more info about the usage code, and the result after using relative path to save an image (e.g. the path info of the image with a screenshot)?

themadmrj:
> Thanks for the PR. Could you provide more info about the usage code, and the result after using relative path to save an image (e.g. the path info of the image with a screenshot)?

There is no other code usage than passing the relative path as the parameter. This is necessary on Android 29 and above because under the DCIM folder, your content might not get references in a sub folder without the relative path.

Here is the code example with screenshots:

**Step1: Gallery of my app before saving the image**

As you can see, there is no subfolder yet. This is the control screenshot.
test1

**Step2: Gallery of my app after saving the image without relativePath**

```
  Future<String> downloadPath() async {
    final name = DateTime.now().microsecondsSinceEpoch ~/
        Duration.microsecondsPerMillisecond;

    String dir;

    if (Platform.isIOS || Platform.isMacOS) {
      dir = (await getApplicationSupportDirectory()).absolute.path;
    } else if (Platform.isAndroid) {
      dir = (await getExternalStorageDirectories(
        type: StorageDirectory.dcim,
      ))![0]
          .absolute
          .path;
    } else {
      dir = (await getDownloadsDirectory())!.absolute.path;
    }

    if (!Directory("$dir/TestFolder").existsSync()) Directory("$dir/TestFolder").createSync();

    return "$dir/TestFolder/test3.jpg";
  }
```

`write image to file success: File: '/storage/emulated/0/Android/data/top.kikt.imagescannerexample/files/DCIM/TestFolder/test3.jpg'`
As you can see, the image is saved and visible under recent, but there is no separate bucket called "TestFolder".
test3


**Step3: Gallery of my app after saving the image WITH relativePath**

```
  Future<String> downloadPath() async {
    final name = DateTime.now().microsecondsSinceEpoch ~/
        Duration.microsecondsPerMillisecond;

    String dir;

    if (Platform.isIOS || Platform.isMacOS) {
      dir = (await getApplicationSupportDirectory()).absolute.path;
    } else if (Platform.isAndroid) {
      dir = (await getExternalStorageDirectories(
        type: StorageDirectory.dcim,
      ))![0]
          .absolute
          .path;
    } else {
      dir = (await getDownloadsDirectory())!.absolute.path;
    }

    if (!Directory("$dir/TestFolder").existsSync()) Directory("$dir/TestFolder").createSync();

    return "$dir/TestFolder/test4.jpg";
  }
```

`final asset = await PhotoManager.editor.saveImageWithPath(file.path, relativePath: "DCIM/TestFolder");`


`write image to file success: File: '/storage/emulated/0/Android/data/top.kikt.imagescannerexample/files/DCIM/TestFolder/test4.jpg'`
As you can see, the image is saved and visible under a separate bucket called "TestFolder".

The relativePath has to start with an Android root folder, either "DCIM" or "Pictures".

lib/src/editor.dart
```diff
@@ -23,36 +23,48 @@ class Editor {
...
+  /// On Android 29 or above, you can use [relativePath] to specify a RELATIVE_PATH used in the MediaStore.
   Future<AssetEntity?> saveImageWithPath(
     String path, {
     String? title,
     String? desc,
+    String? relativePath,
   }) async {
-    return _plugin.saveImageWithPath(path, title: title, desc: desc);
+    return _plugin.saveImageWithPath(path,
+        title: title, desc: desc, relativePath: relativePath);
```
AlexV525, line 51:
```suggestion
    return _plugin.saveImageWithPath(
      path,
      title: title,
      desc: desc,
      relativePath: relativePath,
    );
```

lib/src/editor.dart
```diff
@@ -23,36 +23,48 @@ class Editor {
...
+  /// On Android 29 or above, you can use [relativePath] to specify a RELATIVE_PATH used in the MediaStore.
   Future<AssetEntity?> saveVideo(
     File file, {
     String? title,
     String? desc,
+    String? relativePath,
   }) async {
-    return _plugin.saveVideo(file, title: title, desc: desc);
+    return _plugin.saveVideo(file,
+        title: title, desc: desc, relativePath: relativePath);
```
AlexV525, line 67:
```suggestion
    return _plugin.saveVideo(
      file,
      title: title,
      desc: desc,
      relativePath: relativePath,
    );
```

AlexV525 (commented):
Changes LGTM overall, just some nit picking.

themadmrj:
I added a fix for #463

android/src/main/kotlin/top/kikt/imagescanner/core/utils/Android30DbUtils.kt
```diff
@@ -376,7 +376,14 @@ object Android30DbUtils : IDBUtils {
...
     val timestamp = System.currentTimeMillis() / 1000
 
-    val typeFromStream = URLConnection.guessContentTypeFromStream(inputStream)
+    var typeFromStream: String;
+    if (title.contains(".")){
+      //title contains file extension, form mimeType from it
+      typeFromStream = "image/${File(title).extension}"
+    } else {
+      typeFromStream = URLConnection.guessContentTypeFromStream(inputStream)
+              ?: "image/${File(title).extension}"
```
AlexV525, line 385: Could `.extension` promote the correct type here since title doesn't contain `.` ?

themadmrj: Not really, but I left it since it was there.... ^^
I think if no extension is specified in the title and `guessContentTypeFromStream` returns null, the mimeType should just be set to `image/*`

AlexV525: > the mimeType should just be set to `image/*`

Sure it's definitely better :) .

lib/src/editor.dart
AlexV525: Here still unresolved.

PR #466, opened 2021-03-30
Add modified date for AssetPathEntity

CaiJingLong:


CaiJingLong:
@AlexV525  
Your Feature, you can verify it.

lib/src/filter/filter_options.dart
```diff
@@ -73,6 +82,7 @@ class FilterOptionGroup {
     result["createDate"] = createTimeCond.toMap();
     result["updateDate"] = updateTimeCond.toMap();
     result['containsEmptyAlbum'] = containsEmptyAlbum;
+    result['containsPathModified'] = containsPathModified;
```
AlexV525, line 85: Lack of `this.containsPathModified = other.containsPathModified` in `merge` method.

AlexV525 (changes requested):
Tested. Overall the feature is good to go on Android, but one place still need to be fix.

CaiJingLong:
Regarding the parameter name, do you have any comments?  We can modify it before merging.
Once released, it will trigger a breaking change.

lib/src/filter/filter_options.dart
```diff
@@ -1,5 +1,7 @@
 import 'dart:convert';
 
+import 'package:photo_manager/photo_manager.dart';
```
AlexV525, line 3: Could we import the minimal reference?

AlexV525:
> @AlexV525 Regarding the parameter name, do you have any comments? We can modify it before merging.

Maybe `lastModified`, `updateAt` or `modifiedAt` ? Can't tell which one is better cause IMO they are all the same.

CaiJingLong:
Are there any suggestions for the field in filterOption?

AlexV525:
`FilterOption` looks good. `FilterOptionGroup` seems can be refactored with the constructor, it's `copyWith` method also behaves slightly different from common copy with methods. For example:

```yaml
class FilterOptionGroup {
  FilterOptionGroup({
    FilterOption? imageOption,
    FilterOption? audioOption,
    FilterOption? videoOption,
    this.containsEmptyAlbum = false,
    this.containsPathModified = false,
  });
}
```

Also deprecated methods can be removed.

CaiJingLong:
> `FilterOption` looks good. `FilterOptionGroup` seems can be refactored with the constructor, it's `copyWith` method also behaves slightly different from common copy with methods. For example:
> 
> ```dart
> class FilterOptionGroup {
>   FilterOptionGroup({
>     FilterOption? imageOption,
>     FilterOption? audioOption,
>     FilterOption? videoOption,
>     this.containsEmptyAlbum = false,
>     this.containsPathModified = false,
>   });
> }
> ```
> 
> Also deprecated methods can be removed.

The PR only add new feature.
I will open new pull request to update ctor.

PR #476, opened 2021-04-14
Merge ios and macos code

CaiJingLong:


CaiJingLong:
Compared with the previous version, the macos code has been completely removed.
Instead, use the soft link method to use the ios code.
Use macros to adapt to macos and ios.

CaiJingLong:
In the modification process, the pod library reference method was originally used, but it was later removed because it is not easy to debug the code directly during the development process, which will increase the cost of use.

Issue #452, opened 2021-03-15
[BUG] Pick a video at very first time. AssetEntity file getter return a file instance but the file size keep changing


JimmyTai:
AssetEntity file getter return a file but the size keep changing.

**To Reproduce**
1. Take a video by device camera.
2. Get the file from AssetEntity.
3. Get file length in a loop
4. See file length keep changing.

**Expected behavior**
AssetEntity file getter should return a file after any changes.

**Flutter version**
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 1.22.6, on macOS 11.2.3 20D91 darwin-x64, locale
    zh-Hant-TW)
 
[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.3)
[✓] Xcode - develop for iOS and macOS (Xcode 12.4)
[✓] Android Studio (version 4.1)
[✓] VS Code (version 1.54.2)
[✓] Connected device (1 available)
    ! Error: Chao-Ting’s iPhone is busy: Copying cache files from device. Xcode
      will continue when Chao-Ting’s iPhone is finished. (code -10)

• No issues found!

**Smartphone (please complete the following information):**
 - Device: iPhone 12 Pro max
 - OS: iOS 14.2.1

AlexV525:
The files may downloading from the iCloud

JimmyTai:
@AlexV525

After test file getter and getMediaUrl()  in AssetEntity, they return different file path.

1. The file path from file getter
` /private/var/mobile/Containers/Data/Application/A106C2D9-2CEC-4638-A091-188C533DE4D2/tmp/.video/IMG_0084.MOV`
2. The file path from getMediaUrl()
`/var/mobile/Media/DCIM/100APPLE/IMG_0084.MOV`

It seems getMediaUrl() return the file located in Gallery, but file getter return the file located in App temporary directory.

AlexV525:
Path is a simulated concept on all platforms, since they don't actually provide a real path that can be access directly.

JimmyTai:
After some research, I found fetchFullSizeVideo in PMManager.m will call it self in the progress handler when progress is 1.0. Then it will trigger ExportSession multiple times. After ExportSession create the local file, fetchFullSizeVideo can found the file in local storage and return the file path. But in the in the meantime, ExportSession haven't done the process. Therefore the file size keep changing.

[...]

PR #478, opened 2021-04-16
fetchFullSizeVideo will cause a loop and return file path before export session process done.

JimmyTai:
1. Fix fetchFullSizeVideo flow logic. Avoid keeping call fetchFullSizeVideo when progress is 1.0. This operation cause a loop and return ile path before export session process done. 

2. Change requestAVAssetForVideo to requestExportSessionForVideo. Avoid iCloud file export failure. Ref: https://github.com/banchichen/TZImagePickerController/issues/1073 

3. Add modification date as file name prefix. Avoid to get the cache file if user modified the video.


JimmyTai:
#452

CaiJingLong (approved):
Looks good to me.

Issue #484, opened 2021-04-25
[BUG] Cannot build for platform macos


masewo:
**Describe the bug**
When trying to build an app that includes photo manager 1.1.4 for platform macos then the build fails.

**To Reproduce**
1. Build for platform macos.

**Expected behavior**
App builds.

**Flutter version**
2.0.5

**Log**

```bash
Command CompileSwift failed with a nonzero exit code
Command CompileSwift failed with a nonzero exit code
/Applications/flutter/.pub-cache/hosted/pub.dartlang.org/photo_manager-1.1.4/macos/Classes/core/PMManager.m:510:37: warning: more '%' conversions than data arguments [-Wformat]
      NSLog(@"Asset download fail: %@");
                                   ~^
/Applications/flutter/.pub-cache/hosted/pub.dartlang.org/photo_manager-1.1.4/macos/Classes/core/PMManager.m:585:35: error: type-id cannot have a name
                  resultHandler:^(UIImage *_Nullable image,
                                  ^~~~~~~
/Applications/flutter/.pub-cache/hosted/pub.dartlang.org/photo_manager-1.1.4/macos/Classes/core/PMManager.m:585:43: error: expected ')'
                  resultHandler:^(UIImage *_Nullable image,
                                          ^
/Applications/flutter/.pub-cache/hosted/pub.dartlang.org/photo_manager-1.1.4/macos/Classes/core/PMManager.m:585:34: note: to match this '('
                  resultHandler:^(UIImage *_Nullable image,
                                 ^
/Applications/flutter/.pub-cache/hosted/pub.dartlang.org/photo_manager-1.1.4/macos/Classes/core/PMManager.m:585:33: warning: type specifier missing, defaults to 'int' [-Wimplicit-int]
                  resultHandler:^(UIImage *_Nullable image,
                                ^
/Applications/flutter/.pub-cache/hosted/pub.dartlang.org/photo_manager-1.1.4/macos/Classes/core/PMManager.m:588:7: error: non-void block should return a value
      return;
      ^
/Applications/flutter/.pub-cache/hosted/pub.dartlang.org/photo_manager-1.1.4/macos/Classes/core/PMManager.m:591:57: error: use of undeclared identifier 'info'
    BOOL downloadFinished = [PMManager isDownloadFinish:info];
                                                        ^
[...]

PR #485, opened 2021-04-27
Revert "fetchFullSizeVideo will cause a loop and return file path before export session process done."

CaiJingLong:
Reverts CaiJingLong/flutter_photo_manager#478

Not compatible with macOS.

Related #484 .

### DIFF
--- a/ios/Classes/core/PMManager.m
+++ b/ios/Classes/core/PMManager.m
@@ -473,24 +473,23 @@
                       attributes:@{}
                            error:nil];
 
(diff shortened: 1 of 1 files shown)


CaiJingLong:
Except for PMImageUtil, UIImage/NSImage should not appear in the code, PMImage should be used instead.

The define of PMImage:
```h
#if TARGET_OS_IOS
typedef UIImage PMImage;
#endif

#if TARGET_OS_OSX
typedef NSImage PMImage;
#endif
```

If we want to convert the PMImage to NSData, use the `[PMImageUtil convertToData]` to do it.

JimmyTai:
Sorry about PR #478 . I will try to use PMImage instead of UIImage and check the MacOS compatibility again.

PR #495, opened 2021-05-14
Add permission code for iOS 14

CaiJingLong:


lib/src/type.dart
```diff
@@ -1,3 +1,5 @@
+import 'dart:io';
```
AlexV525, line 1: ?

ios/Classes/PMPlugin.m
```diff
@@ -84,6 +78,46 @@ - (void)onMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result {
...
+    BOOL auth = PHAuthorizationStatusAuthorized == status;
+    [self.manager setAuth:auth];
+    [handler reply:@(status)];
+}
+
+#if TARGET_OS_IOS
+#if __IPHONE_14_0
+
+- (void) handlePermission:(PMManager *)manager handler:(ResultHandler*) handler{
+    if (@available(iOS 14, *)) {
```
lyio, line 91: Is this necessary inside the `#if __IPHONE_14_0` pragma?

yanivshaked:
Hi @CaiJingLong,

I think you should also add support for calling `PHPhotoLibrary.shared().presentLimitedLibraryPicker`, this will invoke a native popup asking the user to add more images to his selection (useful to implement the WhatsApp asset picker pattern).

See [here](https://developer.apple.com/documentation/photokit/phphotolibrary/3616113-presentlimitedlibrarypicker)

Thanks!

CaiJingLong:
> Hi @CaiJingLong,
> 
> I think you should also add support for calling `PHPhotoLibrary.shared().presentLimitedLibraryPicker`, this will invoke a native popup asking the user to add more images to his selection (useful to implement the WhatsApp asset picker pattern).
> 
> See [here](https://developer.apple.com/documentation/photokit/phphotolibrary/3616113-presentlimitedlibrarypicker)
> 
> Thanks!

I don't know much about this mode. Is there any corresponding article / blog that I can refer to?

android/src/main/kotlin/top/kikt/imagescanner/core/entity/PermissionResult.kt
```diff
@@ -0,0 +1,7 @@
+package top.kikt.imagescanner.core.entity
+
+enum class PermissionResult(val value: Int) {
```
AlexV525, line 3: So we don't have `Limited` in this PR?

lib/src/manager.dart
```diff
@@ -4,13 +4,30 @@ Plugin _plugin = Plugin();
...
+    return (await requestPermissionExtend()).isAuth;
+  }
+
+  /// Android: WRITE_EXTERNAL_STORAGE, READ_EXTERNAL_STORAGE, MEDIA_LOCATION
+  ///
+  /// iOS: NSPhotoLibraryUsageDescription of info.plist
+  ///
+  /// macOS of Release.entitlements:
+  ///   - com.apple.security.assets.movies.read-write
+  ///   - com.apple.security.assets.music.read-write
```
AlexV525, line 25:
```suggestion
  ///  - com.apple.security.assets.movies.read-write
  ///  - com.apple.security.assets.music.read-write
```

ios/Classes/PMPlugin.m
```diff
@@ -44,15 +44,9 @@ - (void)onMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result {
   PMManager *manager = self.manager;
```
AlexV525, line 44: Can we implement permissions according to the `permission_handler` plugin? See https://github.com/Baseflow/flutter-permission-handler/blob/master/permission_handler/ios/Classes/strategies/PhotoPermissionStrategy.m

android/src/main/kotlin/top/kikt/imagescanner/core/entity/PermissionResult.kt
lyio: @AlexV525 isn't `limited` an iOS only feature? This is in the Android plugin code, isn't it?

AlexV525: > This is in the Android plugin code, isn't it?

Yes you're right. 👍

CaiJingLong:
@AlexV525 
Do you think the reading and writing of IOS should be separated?

AlexV525:
> Do you think the reading and writing of IOS should be separated?

A separate PR to improve permission handlers might be more appropriate. But regarding the read/write, yes and not only on iOS, the Android's read/write might need to separate too. Let's discuss this later.

yanivshaked:
> I don't know much about this mode. Is there any corresponding article / blog that I can refer to?

https://www.wwdcnotes.com/notes/wwdc20/10641/

https://www.hackingwithswift.com/forums/swiftui/how-to-call-phphotolibrary-s-presentlimitedlibrarypicker-from-swiftui/3747

CaiJingLong:
> > Do you think the reading and writing of IOS should be separated?
> 
> A separate PR to improve permission handlers might be more appropriate. But regarding the read/write, yes and not only on iOS, the Android's read/write might need to separate too. Let's discuss this later.

Read and write permissions are the same permission group in Android.

CaiJingLong:
> > I don't know much about this mode. Is there any corresponding article / blog that I can refer to?
> 
> https://www.wwdcnotes.com/notes/wwdc20/10641/
> 
> https://www.hackingwithswift.com/forums/swiftui/how-to-call-phphotolibrary-s-presentlimitedlibrarypicker-from-swiftui/3747

OK, get it. I'll open new branch to do it.

CaiJingLong:
> > I don't know much about this mode. Is there any corresponding article / blog that I can refer to?
> 
> https://www.wwdcnotes.com/notes/wwdc20/10641/
> 
> https://www.hackingwithswift.com/forums/swiftui/how-to-call-phphotolibrary-s-presentlimitedlibrarypicker-from-swiftui/3747

OK, get it. I'll open new branch to do it.

PR #496, opened 2021-05-15
fix to set cacheOriginBytes

mnlt24:
First, thank you. This package that you developed helps for me.

I've been using this for long time but there has a bug that cannot work of setCacheOriginBytes.
It occurs an exception when calls it.

I fixed and using for myself before but I thought it's better to share this for everyone including you even if it's simple
Thank you

### DIFF
--- a/android/src/main/kotlin/top/kikt/imagescanner/core/PhotoManagerPlugin.kt
+++ b/android/src/main/kotlin/top/kikt/imagescanner/core/PhotoManagerPlugin.kt
@@ -122,10 +122,7 @@ class PhotoManagerPlugin(
         true
       }
       "cacheOriginBytes" -> {
-        val cacheOriginBytes = call.argument<Boolean>("cache")
-        cacheOriginBytes?.let {
-          PhotoManagerPlugin.cacheOriginBytes = cacheOriginBytes
-        }
+        cacheOriginBytes = call.arguments<Boolean>()
         resultHandler.reply(cacheOriginBytes)
         true
       }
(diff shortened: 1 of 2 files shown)


android/src/main/kotlin/top/kikt/imagescanner/core/PhotoManagerPlugin.kt
```diff
@@ -122,8 +122,10 @@ class PhotoManagerPlugin(
         true
       }
       "cacheOriginBytes" -> {
-        val cacheOriginBytes = call.arguments<Boolean>()
-        PhotoManagerPlugin.cacheOriginBytes = cacheOriginBytes
+        val cacheOriginBytes = call.argument<Boolean>("cache")
+        cacheOriginBytes?.let {
+          PhotoManagerPlugin.cacheOriginBytes = cacheOriginBytes
+        }
```
AlexV525, line 128:
```suggestion
        cacheOriginBytes = call.arguments<Boolean>()
```

lib/src/plugin.dart
```diff
@@ -270,7 +270,7 @@ class Plugin with BasePlugin, IosPlugin, AndroidPlugin {
   }
 
   Future<bool> cacheOriginBytes(bool cache) async {
-    return (await _channel.invokeMethod('cacheOriginBytes')) == true;
+    return (await _channel.invokeMethod("cacheOriginBytes", {"cache": cache})) == true;
```
AlexV525, line 273:
```suggestion
    return (await _channel.invokeMethod('cacheOriginBytes', cache)) == true;
```

AlexV525 (approved):
Thanks for the fix!

Issue #455, opened 2021-03-19
[BUG] auto request permission on launch on iOS 12.4


sanekyy:
**Describe the bug**
If I launch app with this package on iOS 12.4, it will ask for permission on app launch.

Reproduced on 12.3 and 12.4. 
iOS 14.+ works correctly.

**Smartphone (please complete the following information):**
 - Device: Any iPhone
 - OS: iOS 12.3-12.4

greymag:
This line 
```
cachingManager = [PHCachingImageManager new];
```
in `PMManager` initiate permissions dialog.

PR #503, opened 2021-05-19
Fixed: permissions dialog on launch on old iOS versions.

greymag:
See #455 

Creation instance of`PHCachingImageManager` initiate request permissions dialog. I made cachingManager lazy to fix it. So now permissions won't be asked on application launch, but only when they really needed.


AlexV525 (approved):
LGTM

AlexV525:
Could you add a description in the CHANGELOG to indicate this change? Thanks.

PR #536, opened 2021-06-24
Add some properties and methods for change notify

CaiJingLong:


CaiJingLong:
And the commits contains an example for the notify of change

Issue #531, opened 2021-06-23
`LateInitializationError` when create `AssetEntityPath`


mahbubmunna:
**Describe the bug**
Creating a AssetEntityPath from id through an error saying lateInitializationError: Field '_type@73180071' has not been initialized.
I want use to the same album in iOS and for that I am saving the id for the further use, but once I'm trying to construct the asset entity path, facing this issue

**To Reproduce**
AssetPathEntity.fromId(myPreviousSavedID);

**iOS Version and Device Info**
iOS version: 14.5
iphone 11 pro Simulator

**Log**

```
[VERBOSE-2:ui_dart_state.cc(199)] Unhandled Exception: LateInitializationError: Field '_type@73180071' has not been initialized.
#0      AssetPathEntity._type (package:photo_manager/src/entity.dart)
#1      AssetPathEntity.type (package:photo_manager/src/entity.dart:52:27)
#2      AssetPathEntity.typeInt (package:photo_manager/src/entity.dart:62:22)
#3      PhotoManager.fetchPathProperties (package:photo_manager/src/manager.dart:237:14)
#4      AssetPathEntity.refreshPathProperties (package:photo_manager/src/entity.dart:89:39)
#5      AssetPathEntity.fromId (package:photo_manager/src/entity.dart:14:18)
#6      _HomePageState.createAlbum (package:auto_wallpaper/screens/home_page.dart:519:48)
<asynchronous suspension>
```

AlexV525:
Please provide a readable log stacktrace with the `_LateInitializationError`, thanks.

mahbubmunna:
> Please provide a readable log stacktrace with the `_LateInitializationError`, thanks.

I have updated issue with full stacktrace

AlexV525:
👍  Thanks, will take a look later.

mahbubmunna:
> 👍 Thanks, will take a look later.

Did you get time to look into, it will be super helpful if you come up with a answer, Thanks

AlexV525:
Can you add a breakpoint at src/entity.dart#L93 to inspect the `result` field ?

mahbubmunna:
Sure, Let me try

mahbubmunna:
I have inspected, before calling src/entity.dart#L93, the call to the function fetchPathProperties in src/entity.dart#L235 is happening, and the result is null
see the attached
[...]

PR #557, opened 2021-07-14
Limit `AssetPathEntity` methods when the permission is not fully authorized

AlexV525:
This pull request intends to fix the root cause of the #531 , that the methods related to the `AssetPathEntity` is not working when the permission is limited, not fully authorized.

### DIFF
--- a/ios/Classes/PMPlugin.m
+++ b/ios/Classes/PMPlugin.m
@@ -181,8 +181,9 @@
 #endif
 }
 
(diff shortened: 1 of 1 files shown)


AlexV525:
Instead of throwing a `FlutterError`, we don't have other choice for now since the interface is limited with the specific type.

CaiJingLong (approved):

PR #559, opened 2021-07-14
Prevent multiple permission requests on Android

### LINKED ISSUE #553: [BUG] Requesting permissions multiple times should throw an exception in dart
**Describe the bug**
When requesting permissions multiple times, flutter_photo_manager currently logs the message `Can request only one set of permissions at a time`. This should instead throw an exception when trying to do this, so that it's more obvious that it's occurring.

The problem is that when requesting permissions multiple times unexpected behavior occurs. If storage access was previously denied, calling `PhotoManager.requestPermissionExtend` or `PhotoManager.requestPermission` multiple times will result in one of the calls instantly returning `PermissionState.authorized` even though it is not authorized and the prompt is being displayed to the user still. This can cause problems in code paths when the permission state gets checked before pushing a page for example.

**To Reproduce**
1. Clone: https://github.com/mhstoller/multiple_permissions_demo
2. Run the app
3. Press the "Request Permissions Once (working)" button
4. Deny storage access (note that `Permission State: PermissionState.denied` is logged after denying access)
5. Press the "Request Permissions Twice (broken)" button
6. Note that while the prompt is still open, `Permission State: PermissionState.authorized` has been logged as well as `Can request only one set of permissions at a time`

**Expected behavior**
`Can request only one set of permissions at a time` should be an exception thrown in Dart, and requesting permissions multiple times shouldn't affect the permission state (ie. if it's denied calling for permissions multiple times shouldn't return authorized if the user hasn't authorized storage permission).

**Smartphone**
Android

AlexV525:
Fix #553.

### DIFF
--- a/android/src/main/kotlin/top/kikt/imagescanner/core/PhotoManagerPlugin.kt
+++ b/android/src/main/kotlin/top/kikt/imagescanner/core/PhotoManagerPlugin.kt
@@ -162,6 +162,11 @@ class PhotoManagerPlugin(
       return
     }
 
+    if (permissionsUtils.isRequesting) {
+      resultHandler.reply(PermissionResult.Denied.value)
+      return
+    }
+
     val utils = permissionsUtils.apply {
       withActivity(activity)
       permissionsListener = object : PermissionsListener {
@@ -169,13 +174,16 @@ class PhotoManagerPlugin(
           LogUtils.info("onDenied call.method = ${call.method}")
           if (call.method == "requestPermissionExtend") {
             resultHandler.reply(PermissionResult.Denied.value)
+            return
+          }
+          if (grantedPermissions.containsAll(arrayListOf(
+                Manifest.permission.READ_EXTERNAL_STORAGE,
+                Manifest.permission.WRITE_EXTERNAL_STORAGE
+            ))) {
+            LogUtils.info("onGranted call.method = ${call.method}")
+            onHandlePermissionResult(call, resultHandler, false)
           } else {
-            if (grantedPermissions.containsAll(arrayListOf(Manifest.permission.READ_EXTERNAL_STORAGE, Manifest.permission.WRITE_EXTERNAL_STORAGE))) {
-              LogUtils.info("onGranted call.method = ${call.method}")
-              onHandlePermissionResult(call, resultHandler, false)
-            } else {
-              replyPermissionError(resultHandler)
-            }
+            replyPermissionError(resultHandler)
           }
         }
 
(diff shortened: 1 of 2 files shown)


mhstoller:
@AlexV525 I tried unsuccessfully to test this.

In my pubspec.yaml I tried:
```
  photo_manager:
    git:
      url: https://github.com/CaiJingLong/flutter_photo_manager
      ref: prevent-multiple-permission-requests
```

and I also tried
```
  photo_manager:
    git:
      url: https://github.com/CaiJingLong/flutter_photo_manager/
      ref: bf51b0b0b440cd170e2a8ca28ae83b7128d86868
```

In both cases after running `flutter pub get` and then debugging the app, the behavior is the same as before. My guess is that it's not using the right package. Do you know what I'm doing wrong, or how I can verify it's pulled the correct version?

AlexV525:
I'll take another look later today.

AlexV525:
Should be good with

```yaml
photo_manager:
  git:
    url: https://github.com/CaiJingLong/flutter_photo_manager/
    ref: 'e4a5bf2'
```

mhstoller:
It does work in one test case, requesting twice when permission was denied does always return denied.

However, requesting twice when permission was approved results in permission always returning denied.

Test flow:
1. Clone: https://github.com/mhstoller/multiple_permissions_demo
2. Run the app
3. Press the "Request Permissions Once (working)" button
4. Allow storage access (note that Permission State: PermissionState. authorized is logged after denying access)
5. Press the "Request Permissions Twice (broken)" button
6. Note that while the prompt is still open, PermissionState.denied has been logged and any subsequent requests for permissions results in PermissionState.denied even though storage is allowed in the app settings

AlexV525:
I'm quite lost with what intention we start this fix. I remember that iOS will return a denied permission and this is to align the behavior between platforms.

So if this is not the thing you expected, is throwing an exception match your expectation?

mhstoller:
On iOS, requesting multiple times always returns the correct permission status. I didn't see issues when I tested it just now.

Ideally, Android would do the same.

The new code causing the issue above is:
```
if (permissionsUtils.isRequesting) {
      resultHandler.reply(PermissionResult.Denied.value)
      return
}
```

I don't think it's appropriate to automatically return denied because the status is unknown and currently being requested. Maybe if it's already requesting, you can just ignore the subsequent requests and not call resultHandler.reply at all.

If it's not possible then an exception should be thrown in Dart so that we don't call for permissions while already waiting for a permission response.

AlexV525:
Again, requesting permission requests twice without waiting is still a mistake.

I'll take the last option, throw an exception when we detects another permission request is ongoing.

AlexV525:
Now it looks like this.

image

mhstoller:
The most beautiful error I've ever seen 😍

I'm good with this solution, thank you!!

AlexV525:
Sounds good. Thank you for keep gave us feedbacks to help us improve our works.

PR #568, opened 2021-07-20
♻️ Make `isLocallyAvailable` as a Future getter

### LINKED ISSUE #567: [BUG] Getting asset list in iOS is slow due to N+1 query
**Describe the bug**

After version 1.2.6, paged, range, and full list (i.e. assetList) calls are very slow to return results. Seems like #560 could be the culprit.

Other developers have complained about this as well. https://developer.apple.com/forums/thread/79296

**Expected behavior**

Fetching assets should always be a fast call by default. Instead of trying to determine if an asset is locally stored while pulling the list, I'd suggest two possible alternatives.

1. An async call from `AssetEntity` that the developer needs to call, which I think would be cleanest.

```
bool isLocal = await assetEntity.isLocallyAvailable();
```
 
2. Adding `isLocallyAvailable` to `PMProgressState` so a developer can access it through a progress handler

AlexV525:
Fixes #567 . /cc @jaysignorello

### DIFF
--- a/ios/Classes/core/PMManager.m
+++ b/ios/Classes/core/PMManager.m
@@ -132,6 +132,19 @@
     return result.count >= 1;
 }
 
+- (BOOL)entityIsLocallyAvailable:(NSString *)assetId {
+    PHFetchResult<PHAsset *> *result =
+    [PHAsset fetchAssetsWithLocalIdentifiers:@[assetId]
+                                     options:[PHFetchOptions new]];
+    if (!result) {
+        return NO;
+    }
+    PHAsset *asset = result.firstObject;
+    NSArray *rArray = [PHAssetResource assetResourcesForAsset:asset];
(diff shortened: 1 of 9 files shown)


ios/Classes/core/PMManager.m
```diff
@@ -132,6 +132,19 @@ - (BOOL)existsWithId:(NSString *)assetId {
     return result.count >= 1;
 }
 
+- (BOOL)entityIsLocallyAvailable:(NSString *)assetId {
+    PHFetchResult<PHAsset *> *result =
+    [PHAsset fetchAssetsWithLocalIdentifiers:@[assetId]
+                                     options:[PHFetchOptions new]];
+    if (!result) {
```
AlexV525, line 139: If the asset does not exist, the result will be false too.

PR #573, opened 2021-07-21
Fix `saveVideo` method return null

### LINKED ISSUE #327: await PhotoManager.editor.saveVideo return `null`
无法使用。这个方法。返回结果为null.无法得到实体

fxfSean:
Fixes #327 . Create the saving directory if it's not exist.


AlexV525:
Thanks for the fix! Did you checked whether other *DBUtils need the same fix?

fxfSean:
> Thanks for the fix! Did you checked whether other *DBUtils need the same fix?

not yet , I just met this problem on this single machine ,and if other *DBUtils has the same way implement, I think it is needed

android/src/main/kotlin/top/kikt/imagescanner/core/utils/DBUtils.kt
```diff
@@ -519,7 +519,12 @@ object DBUtils : IDBUtils {
             inputStream.close()
         } else {
             val tmpPath = assetEntity?.path!!
-            val tmpFile = File(tmpPath)
+            var tmpFile = File(tmpPath)
```
AlexV525, line 522: This fix didn't apply to the `saveVideo` actually, it applied to the `saveImage`.

AlexV525 (changes requested):

android/src/main/kotlin/top/kikt/imagescanner/core/utils/DBUtils.kt
fxfSean: sorry😓 ,I just pasted the code to the wrong place

fxfSean: I just test the `saveImageWithPath` method ,it has the same problem , before apply the code it returns null
image
after apply the code it works ok
image

AlexV525: So it should be applied to every save methods. Thanks for these further tests.

AlexV525:
After all methods been taken care, this PR will be ready for merge.

AlexV525:
> 是否有验证过通过 bytes 保存图片的方法是否存在这个问题？

Did you verify whether the image bytes saving method will occur this problem?

android/src/main/kotlin/top/kikt/imagescanner/core/utils/DBUtils.kt
```diff
@@ -519,7 +519,12 @@ object DBUtils : IDBUtils {
...
+            val directoryFile = File(tmpFile.parent!!)
```
AlexV525, line 523: Can we use `tmpFile.parentFile!!` instead of `File(tmpFile.parent!!)`?

AlexV525:
We should add the directory creation to both `Android30DbUtils` and `AndroidQDBUtils`.

image

fxfSean:
> > 是否有验证过通过 bytes 保存图片的方法是否存在这个问题？
> 
> Did you verify whether the image bytes saving method will occur this problem?

> 通过 bytes 保存图片有同样的问题，我新提交了代码，增加判断

this method `saveImageWithBytes` has the same problem , i pushed the supplement code ,adding judgement logic moment ago

android/src/main/kotlin/top/kikt/imagescanner/core/utils/CommonExt.kt
```diff
@@ -21,3 +22,13 @@ fun useFilePath(): Boolean {
...
+
+/**
+ * Check if the directories exists, if not, create it
+ */
+fun String.checkDirs() {
+    val targetFile = File(this)
+    if (!targetFile.parentFile!!.exists()){
+        targetFile.parentFile!!.mkdirs()
+    }
+}
```
AlexV525, line 34: Need an empty line for EOF.

android/src/main/kotlin/top/kikt/imagescanner/core/utils/DBUtils.kt
```diff
@@ -436,6 +436,7 @@ object DBUtils : IDBUtils {
         cursor.use {
             if (cursor.moveToNext()) {
                 val targetPath = cursor.getString(0)
+                targetPath.checkDirs()
```
AlexV525, line 439: Can you apply the fix to other DBUtils too?

fxfSean: sure, give me a minute

fxfSean: I tested two Android 10 devices, Huawei mateX and Google Pixel3 , and it works ok with the `saveVideo` method, meanwhile the other DBUtils has the different implement, so I think it is no need to apply the fix to them

AlexV525: I think it's necessary because most of feedbacks are using devices with Android 10 and 11, which also require a path check before the insertion.

fxfSean: done.

android/src/main/kotlin/top/kikt/imagescanner/core/utils/CommonExt.kt
```diff
@@ -21,3 +22,13 @@ fun useFilePath(): Boolean {
         else -> true
     }
 }
+
+/**
+ * Check if the directories exists, if not, create it
```
AlexV525, line 27:
```suggestion
 * Create the directory if it's not exist.
```

example/lib/page/save_image_example.dart
```diff
@@ -38,11 +38,12 @@ class _SaveMediaExampleState extends State<SaveMediaExample> {
     if (Platform.isIOS || Platform.isMacOS) {
       dir = (await getApplicationSupportDirectory()).absolute.path;
     } else if (Platform.isAndroid) {
-      dir = (await getExternalStorageDirectories(
-        type: StorageDirectory.downloads,
-      ))![0]
-          .absolute
-          .path;
+      // dir = (await getExternalStorageDirectories(
```
AlexV525, line 41: Remove this?

AlexV525 (commented):
Almost done, just some nit pickings.

example/lib/page/save_image_example.dart
fxfSean: these two `dir` path represent two different way in `DBUtils`->`saveImage` 
```
if (savePath) { //true or false
    inputStream.close()
} else {
 ...
```
so, I suggest keep this

AlexV525: No, I mean the commented out part.

fxfSean: I changed it back, since the `saveVideo` method has acquired `getTemporaryDirectory` path, so it can cover this two cases

AlexV525 (approved):
RSLGTM. Now I think we're covered most of saving failed cases. Thanks for this solid fix! 👍

PR #623, opened 2021-10-18
Improve sort orders on all platforms

AlexV525:
Further fix to #620 , as we found it's better to construct a default `OrderOption` instead of handling separately on different platforms.


CaiJingLong:
Looks good

PR #641, opened 2021-11-10
Include iCloud Shared Albums in the gallery list

### LINKED ISSUE #521: [Custom] Can this access Shared Albums on iOS ?
## Expect
When I retrieve a list of albums on an iOS device I would like to include the users Shared Albums and to be able to view the assets within it.

jaysignorello:
Currently, iCloud Shared Albums are filtered out of the results because Apple returns them as type 
`PHCloudSharedAlbum`, which is inherited from `PHAssetCollection`, but it only allows for exact match of `PHAssetCollection`. This change ensures those albums are returned correctly. 

This should fix #521

### DIFF
--- a/ios/Classes/core/PMManager.m
+++ b/ios/Classes/core/PMManager.m
@@ -108,7 +108,7 @@
         return;
     }
     for (PHCollection *phCollection in collections) {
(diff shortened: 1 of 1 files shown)


AlexV525 (approved):
LGTM. Thanks for the fix!

Issue #652, opened 2021-12-08
[Feature] Better way to handle notify


bdlukaa:
**Is your feature request related to a problem? Please describe.**
When I'm trying to update my media grid when the gallery is updated, I have to handle `MethodCall` by myself. I could figure out how to get the possible values present if `call.arguments`

https://github.com/fluttercandies/flutter_photo_manager/blob/b6885e31e64ea59032e20e029c840dfbd21e6533/android/src/main/kotlin/top/kikt/imagescanner/core/PhotoManagerNotifyChannel.kt#L73-L77

Which makes me do:

```dart
void _onChanged(MethodCall call) {
  final args = call.arguments;
  final path = args['uri'];
  final changeType = args['type'];
  final mediaType = args['mediaType'];
}
```

**Describe the solution you'd like**
A better way to handle it. Something like:

```dart
void _onChanged(MediaChangeInfo info) {
  final path = info.uri;
  final changeType = path.changeType;
  final mediaType = path.mediaType;
}
```

Or maybe even with streams (a more dart-way):

```dart
PhotoManager.onNotify.listen((MediaChangeInfo info) { ... });
```

**Describe alternatives you've considered**
N/A

**Additional context**
- `photo_manager` version `^1.3.10`

AlexV525:
AFAICT, notify events are different between platforms.
https://github.com/fluttercandies/flutter_photo_manager/blob/2ffa0b70926988570520711c3e0292a292041763/ios/Classes/PhotoChangeObserver.m#L35-L37

bdlukaa:
Then, another reason to make a unified version and better way to do so in dart

CaiJingLong:
The problem is that it is difficult to establish uniformly, because Android's notifications about MediaStore actually change frequently. Since API26(The most major API version when this package was created), there have been 3~4 behavior changes.

Although the behavior of iOS has not changed much in recent versions, due to the introduction of limit permissions, the current notifications are also inaccurate.

Moreover, the recent Android should also introduce the limit read permission of the photo album, and may face the same problem as iOS at that time.

Therefore, the notification function may not be able to bring new features for the time being. I plan to make a modification to the notification system after the Android limit permission is completed.

PR #653, opened 2021-12-09
Notify

bdlukaa:
#652 

*ONLY TESTED ON ANDROID*

### Before

```dart
PhotoManager.addChangeCallback(changeNotify);
PhotoManager.startChangeNotify();
```

```dart
PhotoManager.removeChangeCallback(changeNotify);
PhotoManager.stopChangeNotify();
```

### Now

```dart
PhotoManager.onChangeNotify.listen((NotifyChangeInfo info) {
  print(info);
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ you can get assets (image/video/audio) from Android, iOS and macOS.
       - [location info of android Q](#location-info-of-android-q)
       - [Origin description](#origin-description)
       - [Create with id](#create-with-id)
-    - [observer](#observer)
+    - [Observer](#observer)
     - [Clear file cache](#clear-file-cache)
     - [Experimental](#experimental)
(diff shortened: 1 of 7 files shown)


AlexV525 (commented):
I'm not fully reviewed this, but we should move on to iOS/macOS.

bdlukaa:
@AlexV525 I don't have a macos to test if it'd go well. Any suggestions?

AlexV525:
Drafting it, and I'll see if there is any chance to follow this up.

AlexV525:
P.S. It'll be more acceptable if we hold the inner type as `MethodCall`, instead of some unverified abstraction, which allows the main idea of this PR to land first, then tweak them by covering all contents. Also, I would be pleased if new content is well explained and documented, making it easy to review.

Issue #587, opened 2021-08-23
[Feature] Execute stuff on separate thread - or different way to make it faster/not block UI


TheLastGimbus:
Hiii!

When calling any functions from this library, it usually takes *some* time - if I would like to get index of *all* media on device, it can take up to 5 seconds! This often blocks the main UI thread and makes app unresponsive

If there would be a way to move stuff done on native side to some separate thread or something, it would be much better!

AlexV525:
Is this related to all platforms or a specific platform?

TheLastGimbus:
I only have Android phone: Android 10 (LineageOS 17)

TheLastGimbus:
Thank you :heart:  !!!

PR #656, opened 2021-12-21
⚡️ Run Glide on current thread

### LINKED ISSUE #649: [BUG] Android getThumb works in Main thread
Although `getThumb` has `runOnBackground` in `PhotoManagerPlugin`,
`Glide` call `onResourceReady` still work in Main thread.

After that , `resource.compress(format, quality, bos)` may  affect ui performance.

image

The Thread excutor is in main thread.
image

BTW : `getThumbOfUri` may have the same issue, but i don't test yet :(

AlexV525:
Glide will now run on the background thread.

Fix #649, #587.

(I'm not sure if it's truly running on another thread, but it's under the thread pool.)

image


bdlukaa (approved):
LGTM

https://github.com/bumptech/glide/wiki/Loading-and-Caching-on-Background-Threads

Issue #603, opened 2021-09-22
[BUG] Duplicate images on iOS are sorted differently from ios "Photos" app


iosephmagno:
**Describe the bug**
Sometime users duplicate images in their albums (eg. to apply different effects to the same pic). Noticed that all duplicated images are not loaded by photomanager.

**To Reproduce**
1. Load Recents album via photomanager
2. Duplicate an image via native photos app
3. Load album again via photomanager and see that duplicated image is neglected

Tested on iOS, it might affect android as well.

**Expected behavior**
Albums in photomanager should be an exact copy of what they are in native photos app.

**Flutter version**

**Smartphone (please complete the following information):**
 - Device: iPhoneXS
 - OS:  iOS17.7.1

**Log**

If applicable, add logs to help explain your problem.

```bash

```

AlexV525:
Could you tell me how to get a iOS 17 device? I'm really curious about it.

iosephmagno:
Typo :) It’s 14.7.1

AlexV525:
Versions, code snippets, screenshots/screenrecordings, and whether you've granted full permissions or not.

iosephmagno:
How to reproduce issue on iphone XS, iOS 14.7.1, full permissions enabled, video/photos available below:
1. Open wechat assets picker
2. Load recents album
3. Switch to iOS photos and duplicate a photo (not the last one)
4. Switch back to wechat assets picker
5. Load recents album 
6. Duplicate photo in point 3 is ignored in the picker’s grid.

Video: https://www.dropbox.com/s/irws5adivc8exzv/Wechatsassetspicker.MP4?dl=0

Photos:
![828AC7C5-6B1A-4216-AFE2-49B17BBBE353](https://user-images.githubusercontent.com/83366073/134761940-ddc1695c-7495-467c-b027-39dbcfc7dbd5.png)

![8E6CB3CB-2E54-466A-83B8-89AAEC117868](https://user-images.githubusercontent.com/83366073/134761947-5afe2b71-ad81-4f97-82e0-f860d76333b5.png)

![766C28A3-D987-4061-A014-34038DAAB39D](https://user-images.githubusercontent.com/83366073/134761968-88c7751b-bc36-46ea-b281-fe4bff40498a.png)

[...]

PR #659, opened 2021-12-22
🎨 Improve platform default sort orders

AlexV525:
- Flutter: Remove the default `OrderOption` when constructing the group.
- Android: Use `MediaStore.MediaColumns._ID` with descending by default.
- iOS: Use `nil` sort descriptors and reverse the asset array.
  - Suggested by the [answer](https://stackoverflow.com/a/29949179).

### Issues

Related to #655, #603.

### Limitations

- Users won't able to trigger the default sort order manually, only if they pass an empty list of orders.
- Platform handlers are quite separate here, but I left comments, and let's hope it worked.

### DIFF
--- a/ios/Classes/core/PMManager.m
+++ b/ios/Classes/core/PMManager.m
@@ -200,11 +200,9 @@
     }
     
     PHFetchOptions *assetOptions = [self getAssetOptions:type filterOption:filterOption];
-    
     PHAssetCollection *collection = fetchResult.firstObject;
-    
-    PHFetchResult<PHAsset *> *assetArray =
-    [PHAsset fetchAssetsInAssetCollection:collection options:assetOptions];
+    PHFetchResult<PHAsset *> *assetArray = [PHAsset fetchAssetsInAssetCollection:collection
+                                                                         options:assetOptions];
     
     if (assetArray.count == 0) {
         return result;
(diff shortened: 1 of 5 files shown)


ios/Classes/core/PMManager.m
```diff
@@ -222,7 +220,11 @@ - (void)injectAssetPathIntoArray:(NSMutableArray<PMAssetPathEntity *> *)array
     BOOL videoNeedTitle = filterOption.videoOption.needTitle;
     
     for (NSUInteger i = startIndex; i <= endIndex; i++) {
-        PHAsset *asset = assetArray[i];
+        NSUInteger index = i;
+        if (assetOptions.sortDescriptors == nil) {
+            index = count - 1 - i;
```
AlexV525, line 225:
```suggestion
            index = count - i - 1;
```

ios/Classes/core/PMManager.m
```diff
@@ -268,7 +270,11 @@ - (void)injectAssetPathIntoArray:(NSMutableArray<PMAssetPathEntity *> *)array
     }
     
     for (NSUInteger i = startIndex; i <= endIndex; i++) {
-        PHAsset *asset = assetArray[i];
+        NSUInteger index = i;
+        if (assetOptions.sortDescriptors == nil) {
+            index = count - 1 - i;
```
AlexV525, line 275:
```suggestion
            index = count - i - 1;
```

Issue #665, opened 2021-12-24
[BUG]  The Android 30 permission is denied When Android add WRITE_EXTERNAL_STORAGE maxSdkVersion is 29


dyy1993:
**Describe the bug**

The Android 11 permission is denied When Android add WRITE_EXTERNAL_STORAGE maxSdkVersion is 29. (User authorized)

**To Reproduce**
1. add   `<uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"
               android:maxSdkVersion="29" />` in AndroidManifest.xml.
2. PermissionState is denied on 30 Android SDK (User authorized).

**Flutter version**
flutter 2.5.3

**Smartphone (please complete the following information):**
 - OS: [Android 11]

dyy1993:
Google does not recommend using android.permission.WRITE_EXTERNAL_STORAGE in Android 11. [link](https://developer.android.google.cn/training/data-storage/use-cases?hl=zh-cn#show-all-files-media)

AlexV525:
It didn't tell *NOT RECOMMEND* on the doc, but another behavior was provided instead.

When permission is only used on SDKs below 29, then it should be at the denied state on 29+, isn't it correct?

dyy1993:
> It didn't tell _NOT RECOMMEND_ on the doc, but another behavior was provided instead.
> 
> When permission is only used on SDKs below 29, then it should be at the denied state on 29+, isn't it correct?

When setting <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" android:maxSdkVersion="29" /> in other plugins, this plugin checkPermissions is false in 29+.
Please review this pr #666.

dyy1993:
For example, other plugin add this.
![WX20211224-184601](https://user-images.githubusercontent.com/10572649/147346311-3f98789d-3a7a-489a-88a6-82ccaea79d97.png)
photo manage not working.

PR #666, opened 2021-12-24
🚑  fix Android11 permission bug

dyy1993:
### Type of the change

* [ ] Feature
* [x] Bugfix
* [ ] Technical
* [ ] Documentation
* [ ] Other

### Description

The Android 11 permission is denied When Android add WRITE_EXTERNAL_STORAGE maxSdkVersion is 29. (User authorized)
#665


AlexV525:
What if the `WRITE_EXTERNAL_STORAGE` is truly denied on the current device?

AlexV525:
If we can't figure out what exactly the problem is, this PR is still having false positives when handling permissions.

Converting this to draft since we don't have a clear conclusion here.

dyy1993:
> If we can't figure out what exactly the problem is, this PR is still having false positives when handling permissions.
> 
> Converting this to draft since we don't have a clear conclusion here.

We refer to permission_handler to reprocess the write permission on Android 11. [permission-handler pr](https://github.com/Baseflow/flutter-permission-handler/pull/526)
WechatIMG236

AlexV525:
Again with https://github.com/fluttercandies/flutter_photo_manager/issues/665#issuecomment-1000770787, they're new recommended methods rather than required methods. You're indicating various scopes for your request. Please make them clear.

AlexV525:
#675

PR #716, opened 2022-02-28
Fix activity leaks on Android

yrom:
Release reference of Activity on PhotoManagerPlugin detached from Activity


AlexV525 (approved):
LGTM. Thanks for the fix!

android/src/main/kotlin/com/fluttercandies/photo_manager/PhotoManagerPlugin.kt
```diff
@@ -50,6 +50,9 @@ class PhotoManagerPlugin : FlutterPlugin, ActivityAware {
         binding?.let {
             onRemoveRequestPermissionResultListener(it)
         }
+        // release Activity instance reference on detached
```
AlexV525, line 53:
```suggestion
        // Release the Activity reference on detached.
```

PR #717, opened 2022-02-28
Support 'mimeType' for iOS and macOS

yrom:
On iOS, we can detecting MIME type from UTI.
See https://developer.apple.com/documentation/photokit/phassetresource/1623989-uniformtypeidentifier?language=objc and https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/understanding_utis/understand_utis_conc/understand_utis_conc.html#//apple_ref/doc/uid/TP40001319-CH202-SW1

### DIFF
--- a/ios/Classes/core/PHAsset+PM_COMMON.m
+++ b/ios/Classes/core/PHAsset+PM_COMMON.m
@@ -4,6 +4,11 @@
 //
 
 #import "PHAsset+PM_COMMON.h"
+#if TARGET_OS_IOS || TARGET_OS_WATCH || TARGET_OS_TV
+#import <MobileCoreServices/MobileCoreServices.h>
+#else
+#import <CoreServices/CoreServices.h>
+#endif
 
 @implementation PHAsset (PM_COMMON)
 
@@ -88,6 +93,15 @@
         
         return @"";
     }
+}
+
+- (NSString *)mimeType {
+    PHAssetResource *resource = [[PHAssetResource assetResourcesForAsset:self] firstObject];
+    if (resource) {
+        NSString *uti = resource.uniformTypeIdentifier;
+        return (__bridge_transfer NSString *)UTTypeCopyPreferredTagWithClass((__bridge CFStringRef)uti, kUTTagClassMIMEType);
+    }
+    return nil;
 }
 
 - (BOOL)isAdjust {
--- a/ios/Classes/core/PHAsset+PM_COMMON.m
+++ b/ios/Classes/core/PHAsset+PM_COMMON.m
@@ -94,7 +94,7 @@
         return @"";
     }
 }
-// UTI:  https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/understanding_utis/understand_utis_intro/understand_utis_intro.html#//apple_ref/doc/uid/TP40001319
+// UTI: https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/understanding_utis/understand_utis_intro/understand_utis_intro.html#//apple_ref/doc/uid/TP40001319
 - (NSString *)mimeType {
     PHAssetResource *resource = [[PHAssetResource assetResourcesForAsset:self] firstObject];
     if (resource) {
--- a/ios/Classes/core/PHAsset+PM_COMMON.m
+++ b/ios/Classes/core/PHAsset+PM_COMMON.m
@@ -94,7 +94,7 @@
         return @"";
     }
 }
-
+// UTI:  https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/understanding_utis/understand_utis_intro/understand_utis_intro.html#//apple_ref/doc/uid/TP40001319
 - (NSString *)mimeType {
     PHAssetResource *resource = [[PHAssetResource assetResourcesForAsset:self] firstObject];
     if (resource) {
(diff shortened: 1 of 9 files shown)


ios/Classes/core/PMConvertUtils.m
```diff
@@ -101,6 +101,7 @@ + (NSDictionary *)convertPMAssetToMap:(PMAssetEntity *)asset
         @"lat": @(asset.lat),
         @"title": needTitle ? asset.title : @"",
         @"subtype": @(asset.subtype),
+        @"mimeType": asset.mimeType ?: [NSNull null],
```
AlexV525, line 104: Is the explicitly `NSNull` necessary?

AlexV525, line 104: Is the explicitly `NSNull` necessary?

ios/Classes/core/PMManager.m
```diff
@@ -332,6 +337,13 @@ - (PMAssetEntity *)convertPHAssetToAssetEntity:(PHAsset *)asset
     entity.title = needTitle ? [asset title] : @"";
     entity.favorite = asset.isFavorite;
     entity.subtype = asset.unwrappedSubtype;
+
+    PHAssetResource *resource = [[PHAssetResource assetResourcesForAsset:asset] firstObject];
```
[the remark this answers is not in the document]
yrom: Use property `mimeTypeAsync` instead

lib/src/types/entity.dart
```diff
@@ -699,12 +699,20 @@ class AssetEntity {
 
   /// The mime type of the asset.
   ///  * Android: `MediaStore.MediaColumns.MIME_TYPE`.
-  ///  * iOS/macOS: Always null.
+  ///  * iOS/macOS: Always null. Use the async getter [mimeTypeAsync] instead.
   ///
   /// See also:
   ///  * https://developer.android.com/reference/android/provider/MediaStore.MediaColumns#MIME_TYPE
```
AlexV525, line 705: Add doc references for both methods.

```suggestion
  ///  * [mimeTypeAsync] which is the asynchronized getter of the MIME type.
  ///  * https://developer.android.com/reference/android/provider/MediaStore.MediaColumns#MIME_TYPE
```

lib/src/types/entity.dart
```diff
@@ -699,12 +699,20 @@ class AssetEntity {
...
   final String? mimeType;
 
+  /// Get the mime type async.
+  ///  * Android: `MediaStore.MediaColumns.MIME_TYPE`.
+  ///  * iOS/macOS: MIME type from `PHAssetResource.uniformTypeIdentifier`.
+  /// 
+  /// See also:
```
AlexV525, line 712: Same as above

AlexV525 (commented):
More thoughts

ios/Classes/core/PMManager.m
```diff
@@ -1077,6 +1077,15 @@ - (NSString *)getTitleAsyncWithAssetId:(NSString *)assetId {
     return @"";
 }
 
+- (NSString *)getMimeTypeAsyncWithAssetId:(NSString *)assetId {
+    PHFetchResult *fetchResult = [PHAsset fetchAssetsWithLocalIdentifiers:@[assetId] options:nil];
+    PHAsset *asset = [self getFirstObjFromFetchResult:fetchResult];
```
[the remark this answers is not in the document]
yrom: IMHO, Live Photo is a special variant of image type. However, does Live Photo require another property like `mimeTypes`?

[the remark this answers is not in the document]
AlexV525: According to https://developer.apple.com/documentation/uniformtypeidentifiers/uttype/3551525-livephoto, it has its own UTI format.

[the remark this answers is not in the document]
yrom: However, `UTType.livePhoto.preferredMIMEType` will return you a `nil`. Since a Live Photo is actually two files, there is no MIME Type for it.

ios/Classes/core/PHAsset+PM_COMMON.m
```diff
@@ -89,6 +94,15 @@ - (NSString *)title {
         return @"";
     }
 }
+// UTI:  https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/understanding_utis/understand_utis_intro/understand_utis_intro.html#//apple_ref/doc/uid/TP40001319
```
AlexV525, line 97:
```suggestion
// UTI: https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/understanding_utis/understand_utis_intro/understand_utis_intro.html#//apple_ref/doc/uid/TP40001319
```

lib/src/types/entity.dart
```diff
@@ -699,12 +699,24 @@ class AssetEntity {
...
   final String? mimeType;
 
+  /// Get the mime type async.
+  ///  * Android: `MediaStore.MediaColumns.MIME_TYPE`.
+  ///  * iOS/macOS: MIME type from `PHAssetResource.uniformTypeIdentifier`.
+  /// 
+  /// For Live Photos on iOS/macOs, this returns a type representing its image file.
+  /// 
+  /// See also:
+  ///  * [mimeType] the synchronized getter of the MIME type
```
AlexV525, line 716:
```suggestion
  ///  * [mimeType] which is the synchronized getter of the MIME type.
```

AlexV525 (approved):
LGTM with nits

Issue #739, opened 2022-03-23
[BUG] Unable to access specified image title after saving it to the gallery


PhoenixPhighter:
**Describe the bug**
I have a project where in one Dart file, I am saving a network image to the iOS gallery with a specified title (in my case, a hashcode). In another Dart file, I need to access that title that I specified, but widget.asset.title and widget.asset.titleAsync are not returning the title I set. Let's say when I am saving the network image, I set the title to be "55775351". In another screen, which displays asset entities from the iOS gallery, when I return the title, I get something like "IMG_3430" instead of the title I specified.

**To Reproduce**

1. Save a network image with a specific title.
final AssetEntity? entity = await PhotoManager.editor.saveImage(rawData, title: hashString + ".png");

2. Print the ID of this AssetEntity and you will see that it is not the specified title, even though in the camera roll you can see the title you specified.
String s = await entity.titleAsync;
print(s);

**Expected behavior**
Returning the Asset title should give the string you specified when you saved it. If you look in the iOS device gallery (since iOS now lets you swipe up and see metadata), you can see that the title you specified is being saved. But, if you try to retrieve that title, it outputs something different.

**Screenshots**
https://imgur.com/a/LAoryLe 
In the screenshot from my iOS camera gallery, you can see that the title is 55775351, but when I fetch the title for this asset, it returns "IMG_3430" (this title follows sequentially from previous photos, so internally iOS is somehow rewriting it I guess - still, the title that I set displays, so should I not be able to access it?)

**Flutter version**
Flutter Version 2.10.3

**Smartphone (please complete the following information):**
 - Device: iPhone 12 Pro Max
 - OS: 15.3
 - Browser: Doesn't matter here, but Firefox

I should note, I read a similar issue where someone was having a similar problem, where they set a path but the newly created AssetEntity had an empty title - basically, the title they set was not being written in. This issue was marked as CLOSED but I could not find any solution in that thread. If you can, instead of just linking that issue since I already read it, could you please explain how to fix this? Thank you.

CaiJingLong:
The title will be use in next code:

https://github.com/fluttercandies/flutter_photo_manager/blob/a8c193ed320bed5e833fcb7a93998e1652f9fec8/ios/Classes/core/PMManager.m#L1006

[The apple document](https://developer.apple.com/documentation/photokit/phassetresourcecreationoptions/1622683-originalfilename?language=objc)

Description in the official document: this field is the original name of the resource.

[...]

PR #740, opened 2022-03-23
Modify the titleAsync implementation on iOS.

CaiJingLong:
Related #739

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## 2.0.5
+
+New behavior:
+
+- Modify the `titleAsync` of `AssetEntity` implementation on iOS. (#740)
+
 ## 2.0.4
 
 Fixes:
--- a/ios/Classes/core/PHAsset+PM_COMMON.h
+++ b/ios/Classes/core/PHAsset+PM_COMMON.h
@@ -18,6 +18,9 @@ NS_ASSUME_NONNULL_BEGIN
 - (int)unwrappedSubtype;
 
 - (NSString*)title;
+
+- (NSString *)originalFileName;
+
 /**
  Get the MIME type for this asset from UTI (`PHAssetResource.uniformTypeIdentifier`), such as `image/jpeg`, `image/heic`, `video/quicktime`, etc.
  
--- a/ios/Classes/core/PHAsset+PM_COMMON.h
+++ b/ios/Classes/core/PHAsset+PM_COMMON.h
@@ -19,7 +19,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (NSString*)title;
 
-- (NSString *)originalFileNameWithSubtype:(BOOL) subtype;
+- (NSString *)originalFileNameWithSubtype:(int) subtype;
 
 /**
  Get the MIME type for this asset from UTI (`PHAssetResource.uniformTypeIdentifier`), such as `image/jpeg`, `image/heic`, `video/quicktime`, etc.
--- a/ios/Classes/core/PHAsset+PM_COMMON.h
+++ b/ios/Classes/core/PHAsset+PM_COMMON.h
@@ -19,7 +19,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (NSString*)title;
 
-- (NSString *)originalFileNameWithSubtype:(int) subtype;
+- (NSString *)originalFileNameWithSubtype:(BOOL) subtype;
 
 /**
  Get the MIME type for this asset from UTI (`PHAssetResource.uniformTypeIdentifier`), such as `image/jpeg`, `image/heic`, `video/quicktime`, etc.
--- a/ios/Classes/core/PHAsset+PM_COMMON.h
+++ b/ios/Classes/core/PHAsset+PM_COMMON.h
@@ -19,7 +19,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (NSString*)title;
 
-- (NSString *)originalFileName;
+- (NSString *)originalFileNameWithSubtype:(int) subtype;
 
 /**
  Get the MIME type for this asset from UTI (`PHAssetResource.uniformTypeIdentifier`), such as `image/jpeg`, `image/heic`, `video/quicktime`, etc.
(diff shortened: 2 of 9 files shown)


lib/src/types/entity.dart
```diff
@@ -343,7 +343,7 @@ class AssetEntity {
   final String? title;
 
   ///  * Android: `MediaStore.MediaColumns.DISPLAY_NAME`.
-  ///  * iOS/macOS: `PHAssetResource.filename`.
+  ///  * iOS/macOS: [`PHAssetResource.originalFilename`](https://developer.apple.com/documentation/photokit/phassetresource/1623985-originalfilename?language=objc).
```
AlexV525, line 346: We don't really need the link though.

ios/Classes/core/PHAsset+PM_COMMON.m
```diff
@@ -94,6 +94,26 @@ - (NSString *)title {
         return @"";
     }
 }
+
+- (NSString *)originalFileName {
```
AlexV525, line 98: Now since we support subtypes, could we also support obtaining the title from the resource that matched the subtype? Also, the `titleAsync` might have an argument like `bool withSubtype = false`.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## 2.0.5
+
+New behavior:
+
+- Modify the `titleAsync` of `AssetEntity` implementation on iOS. (#740)
```
AlexV525, line 7:
```suggestion
Improvements:
- Improve `AssetEntity.titleAsync`'s implementation on iOS. (#740)
```

AlexV525 (changes requested):

ios/Classes/core/PHAsset+PM_COMMON.m
CaiJingLong: The type is PHAssetResourceType, not PHAssetMediaType

CaiJingLong: https://developer.apple.com/documentation/photokit/phassetresourcetype?language=objc

lib/src/types/entity.dart
CaiJingLong:
```suggestion
  ///  * iOS/macOS: `PHAssetResource.originalFilename`.
```

ios/Classes/core/PHAsset+PM_COMMON.m
CaiJingLong: If we need to get the originalFileName based on a different resource type, may need to return a Map<ResourceType, String>

AlexV525: > The type is PHAssetResourceType, not PHAssetMediaType

Indeed. But when we fetch videos for live photos, we first check the `PHAssetMediaType` (subtype), then fetch resources accordingly.

https://github.com/fluttercandies/flutter_photo_manager/blob/a8c193ed320bed5e833fcb7a93998e1652f9fec8/ios/Classes/core/PMManager.m#L422-L423

https://github.com/fluttercandies/flutter_photo_manager/blob/a8c193ed320bed5e833fcb7a93998e1652f9fec8/ios/Classes/core/PMManager.m#L445-L446

CaiJingLong: like this
image

ios/Classes/core/PHAsset+PM_COMMON.m
```diff
@@ -94,6 +94,41 @@ - (NSString *)title {
...
+            }
+        }
+    }
+    
+    PHAssetResource *firstRes = array.firstObject;
+    if (firstRes) {
+        return firstRes.originalFilename;
+    }
+    
+    return @"";
```
AlexV525, line 129:
```suggestion
    BOOL isLivePhoto = [self isLivePhoto] && subtype == PHAssetMediaSubtypePhotoLive;
    if (@available(iOS 9.1, *) || @available(macOS 10.11, *)) {
        if (isLivePhoto) {
            return [self getLivePhotosResource].originalFilename;
        }
    }
    PHAssetResource *resource = [self getAdjustResource];
    if (resource) {
        return resource.originalFilename;
    }
    return @"";
```

ios/Classes/core/PHAsset+PM_COMMON.m
```diff
@@ -94,6 +94,41 @@ - (NSString *)title {
         return @"";
     }
 }
+
+- (NSString *)originalFileNameWithSubtype:(int)subtype {
```
AlexV525, line 98:
```suggestion
- (NSString *)originalFilename:(int)subtype {
```

ios/Classes/core/PHAsset+PM_COMMON.h
```diff
@@ -18,6 +18,9 @@ NS_ASSUME_NONNULL_BEGIN
 - (int)unwrappedSubtype;
 
 - (NSString*)title;
+
+- (NSString *)originalFileNameWithSubtype:(int) subtype;
```
AlexV525, line 22:
```suggestion
- (NSString *)originalFilename:(int) subtype;
```

ios/Classes/core/PMManager.m
```diff
@@ -1079,10 +1079,10 @@ - (void)saveVideo:(NSString *)path
     }];
 }
 
-- (NSString *)getTitleAsyncWithAssetId:(NSString *)assetId {
+- (NSString *)getTitleAsyncWithAssetId:(NSString *)assetId subtype:(int)subtype {
     PHAsset *asset = [PHAsset fetchAssetsWithLocalIdentifiers:@[assetId] options:nil].firstObject;
     if (asset) {
-        return [asset title];
+        return [asset originalFileNameWithSubtype: subtype];
```
AlexV525, line 1085:
```suggestion
        return [asset originalFilename:subtype];
```

AlexV525 (changes requested):

ios/Classes/core/PHAsset+PM_COMMON.m
CaiJingLong: @available This is not valid.

ios/Classes/core/PHAsset+PM_COMMON.m
CaiJingLong: It is a general guideline in objective-c to have a method signature with with arguments.

AlexV525 (approved):

PR #743, opened 2022-03-25
Fix ios clear cache

### LINKED ISSUE #741: [BUG] Bug in clearFileCache() on iOS?
I've noticed that my "Documents & Data" Size on my iPhone is getting bigger and bigger, so I looked into it and found around ~1000 *_exif.jpg Files in my AppContainer.
That is weird because I am using only 100 Images at a time and I always do a `.delete()` on iOS after using `.file` or `.orginFile` and on every app start I also do an additional `.clearFileCache()`

I had that problem already and fixed it with `.clearFileCache()` and `.delete()`, but now it seems to be back.
Shouldn't `.clearFileFache()` delete everything cached?
When starting the app I see it disappearing from the iPhone storage Information, so something is happening there, but when it reappears it's still large 

I am using the latest version of flutter_photo_manager (2.0.5, updated today) and iOS (15.4)

CaiJingLong:
closed #741

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 ## 2.0.6
 
-Fixeds:
+Fixes:
 
-- Clear file cache on iOS. (#742)(#743)
+- Fix file caches clearing on iOS. (#743)
 
 ## 2.0.5
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,18 +1,27 @@
 # CHANGELOG
 
+## 2.0.6
+
+Fixeds:
+
+- Clear file cache on iOS. (#742)(#743)
+
 ## 2.0.5
 
 Improvements:
+
 - Improve `AssetEntity.titleAsync`'s implementation on iOS. (#740)
 
 ## 2.0.4
 
 Fixes:
+
 - Fix invalid `InputStream` when saving images on Android. (#736)
 
 ## 2.0.3
 
 Improvements:
+
 - Improve `getMediaUrl` on iOS.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,33 +1,45 @@
 # CHANGELOG
 
+## 2.0.6
+
+Fixeds:
```
AlexV525, line 5:
```suggestion
Fixes:
```

CHANGELOG.md
```diff
@@ -1,33 +1,45 @@
...
+
+- Clear file cache on iOS. (#742)(#743)
```
AlexV525, line 7:
```suggestion
- Fix file caches clearing on iOS. (#743)
```

ios/Classes/core/PMManager.h
```diff
@@ -16,8 +16,13 @@ typedef void (^ChangeIds)(NSArray<NSString *> *);
 #import "PMResultHandler.h"
 #import "PMConvertProtocol.h"
 
+#define PM_VIDEO_CACHE_PATH @".video"
+#define PM_IMAGE_CACHE_PATH @".image"
+#define PM_FULL_IMAGE_CACHE_PATH @"flutter-images"
```
AlexV525, line 21: Should we use `pm-images`?

CaiJingLong: There is a historical reason here, previously `flutter-images` was used, if it is changed to `pm-images` here, the old images will not be deleted in method.

AlexV525 (approved):

PR #772, opened 2022-05-26
[iOS] Predicate for live-photo filtering workaround

yu840915:
Use `NOT` instead of `!=` in live-photo filtering expression.
With original predicate, a lot of non-live-photo cannot bet fetched unexpectedly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Improvements:
 
 - Correct `PermissionRequestOption` typo with a class type alias.
   Which also raised the Dart SDK constraint to `2.13.0`.
+- The original predicate for filtering live-photo cannot be executed correctly and unexpectedly rejects photos with media subtype `PHAssetMediaSubtypeNone`. Changing it from `( ( mediaSubtype & 8 ) != 8 )` to `NOT ( ( mediaSubtype & 8 ) == 8` works correctly.
 
 ## 2.1.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Improvements:
 
 - Correct `PermissionRequestOption` typo with a class type alias.
   Which also raised the Dart SDK constraint to `2.13.0`.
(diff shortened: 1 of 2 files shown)


AlexV525:
> With original predicate, a lot of non-live-photo cannot bet fetched unexpectedly.

Can you explain more with "cannot be fetched unexpectedly"?

yu840915:
### Before:
On iOS, `PhotoManager.getAssetPathList(type: RequestType.image, filterOption: FilterOptionGroup(containsLivePhotos: false))`, returns number of albums much fewer than native **Photos** app has.

### After: 
`PhotoManager.getAssetPathList(type: RequestType.image, filterOption: FilterOptionGroup(containsLivePhotos: false))`, returns number albums matches native **Photos** app.

AlexV525:
> returns number of albums much fewer than native **Photos** app has.

Why it's fewer before? What's the final predicate before and after applying the fix? How does it affect photo fetching? What kinds of photos are prevented to be fetched from the previous predicate?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Improvements:
 
 - Correct `PermissionRequestOption` typo with a class type alias.
   Which also raised the Dart SDK constraint to `2.13.0`.
+- The original predicate for filtering live-photo cannot be executed correctly and unexpectedly rejects photos with media subtype `PHAssetMediaSubtypeNone`. Changing it from `( ( mediaSubtype & 8 ) != 8 )` to `NOT ( ( mediaSubtype & 8 ) == 8` works correctly.
```
AlexV525, line 13: We don't need a detailed explanation in the CHANGELOG. It should be placed around the updated code instead.

AlexV525 (changes requested):

CHANGELOG.md
yu840915: Agree, how about this?
```suggestion
- The original predicate for filtering live-photo cannot be executed correctly and unexpectedly rejects photos with media subtype `PHAssetMediaSubtypeNone`. Changing it from `( ( mediaSubtype & 8 ) != 8 )` to `NOT ( ( mediaSubtype & 8 ) == 8` works correctly.
- Workaround for live-photo filtering bug in PhotoAsset framework on iOS.
```

AlexV525:
```suggestion
- Improve Live-Photos filtering.
```

PR #786, opened 2022-07-04
[Android] Add `forceGetLocation` filed to get AssetEntity

lantah-1:
Use scenario: When retrieving List<Asset Entity> object, you want to return the location information of all Asset Entity objects directly, instead of iterating through the List to retrieve it individually through get Lat Lng Async method, which is time-consuming

使用场景：获取List<AssetEntity>对象时，想要直接返回所有AssetEntity对象的地理位置信息，而不是遍历list去单独的通过getLatLngAsync方法去获取，那样很耗时。

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,5 +15,3 @@ fvm
 doc/
 .env
 *.iml
-
-.fvm
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ fvm
 doc/
 .env
 *.iml
+
+.fvm
--- a/android/src/main/kotlin/com/fluttercandies/photo_manager/core/entity/FilterOption.kt
+++ b/android/src/main/kotlin/com/fluttercandies/photo_manager/core/entity/FilterOption.kt
@@ -30,6 +30,7 @@ class FilterCond {
     var isShowTitle = false
     lateinit var sizeConstraint: SizeConstraint
     lateinit var durationConstraint: DurationConstraint
+    var isNeedGeo = false
 
     companion object {
         private const val widthKey = MediaStore.Files.FileColumns.WIDTH
--- a/android/src/main/kotlin/com/fluttercandies/photo_manager/core/utils/AndroidQDBUtils.kt
+++ b/android/src/main/kotlin/com/fluttercandies/photo_manager/core/utils/AndroidQDBUtils.kt
@@ -148,8 +148,7 @@ object AndroidQDBUtils : IDBUtils {
         page: Int,
         size: Int,
         requestType: Int,
-        option: FilterOption,
-        forceGetLocation: Boolean
+        option: FilterOption
     ): List<AssetEntity> {
         val isAll = galleryId.isEmpty()
         val list = ArrayList<AssetEntity>()
@@ -175,9 +174,10 @@ object AndroidQDBUtils : IDBUtils {
             args.toTypedArray(),
             sortOrder
         ) ?: return emptyList()
+        val needGeo = if (requestType == 1) option.imageOption.isNeedGeo else false
         cursor.use {
             cursorWithRange(it, page * size, size) {
-                val asset = convertCursorToAssetEntity(context, cursor, forceGetLocation)
+                val asset = convertCursorToAssetEntity(context, cursor, needGeo)
                 list.add(asset)
             }
         }
@@ -191,8 +191,7 @@ object AndroidQDBUtils : IDBUtils {
         start: Int,
         end: Int,
         requestType: Int,
-        option: FilterOption,
-        forceGetLocation: Boolean
+        option: FilterOption
     ): List<AssetEntity> {
         val isAll = galleryId.isEmpty()
         val list = ArrayList<AssetEntity>()
@@ -219,9 +218,10 @@ object AndroidQDBUtils : IDBUtils {
             args.toTypedArray(),
             sortOrder
         ) ?: return emptyList()
+        val needGeo = if (requestType == 1) option.imageOption.isNeedGeo else false
         cursor.use {
             cursorWithRange(it, start, pageSize) {
-                list.add(convertCursorToAssetEntity(context, cursor, forceGetLocation))
+                list.add(convertCursorToAssetEntity(context, cursor, needGeo))
             }
         }
         return list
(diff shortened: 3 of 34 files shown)


.gitignore
```diff
@@ -15,3 +15,5 @@ fvm
 doc/
 .env
 *.iml
+
+.fvm
```
AlexV525, line 19: Remove this.

android/src/main/kotlin/com/fluttercandies/photo_manager/core/utils/IDBUtils.kt
```diff
@@ -94,9 +94,10 @@ interface IDBUtils {
         size: Int,
         requestType: Int = 0,
         option: FilterOption,
+        forceGetLocation: Boolean
     ): List<AssetEntity>
 
-    fun getAssetEntity(context: Context, id: String): AssetEntity?
+    fun getAssetEntity(context: Context, id: String,forceGetLocation: Boolean = false): AssetEntity?
```
AlexV525, line 100: `forceGetLocation` should be inside the `FilterOption`. Also, the code is not well-formatted.

android/src/main/kotlin/com/fluttercandies/photo_manager/core/utils/AndroidQDBUtils.kt
```diff
@@ -248,7 +254,9 @@ object AndroidQDBUtils : IDBUtils {
         val modifiedDate = cursor.getLong(MediaStore.MediaColumns.DATE_MODIFIED)
         val orientation: Int = cursor.getInt(MediaStore.MediaColumns.ORIENTATION)
         val relativePath: String = cursor.getString(MediaStore.MediaColumns.RELATIVE_PATH)
-        if ((width == 0 || height == 0)
+        var lat = 0.0
+        var log = 0.0
+        if ((width == 0 || height == 0 || forceGetLocation)
```
AlexV525, line 259: The request of `ExifInterface.latLong` breaks the condition of obtaining the `width` and `height`.

AlexV525 (changes requested):

AlexV525:
The PR itself seems ineffective. EXIF should depend on permissions on recent Android versions, and the request will make the loading complexity increase from O(n) to O(n^2).

android/src/main/kotlin/com/fluttercandies/photo_manager/core/utils/AndroidQDBUtils.kt
lantah-1: Suppose there are 100,000 pictures in the album, and the pictures with geographic location information need to be found out from them. According to the previous logic, 100,000 pictures need to be recycled first, and then 100,000 times to obtain the exIF of the picture. Now, the need Geo field is used to control the extraction of exIF information when the picture is taken out, which saves the cycle time

假设相册有10万张图片，需要从其中找出有地理位置信息的图片，按照之前的逻辑就需要先循环取出10万张图片，再循环10万次去获取图片的exif,而现在通过needGeo字段去控制在取出图片的同时取出exif信息，节省了循环的时间

lantah-1:
> The PR itself seems ineffective. EXIF should depend on permissions on recent Android versions, and the request will make the loading complexity increase from O(n) to O(n^2).

Suppose there are 100,000 pictures in the album, and the pictures with geographic location information need to be found out from them. According to the previous logic, 100,000 pictures need to be recycled first, and then 100,000 times to obtain the exIF of the picture. Now, the need Geo field is used to control the extraction of exIF information when the picture is taken out, which saves the cycle time

假设相册有10万张图片，需要从其中找出有地理位置信息的图片，按照之前的逻辑就需要先循环取出10万张图片，再循环10万次去获取图片的exif,而现在通过needGeo字段去控制在取出图片的同时取出exif信息，节省了循环的时间

AlexV525:
> Suppose there are 100,000 pictures in the album, and the pictures with geographic location information need to be found out from them. According to the previous logic, 100,000 pictures need to be recycled first, and then 100,000 times to obtain the exIF of the picture. Now, the need Geo field is used to control the extraction of exIF information when the picture is taken out, which saves the cycle time

This only saves callers, not the total time. Also, I don't see you're handling the permission here.

android/src/main/kotlin/com/fluttercandies/photo_manager/core/entity/FilterOption.kt
```diff
@@ -30,6 +30,7 @@ class FilterCond {
     var isShowTitle = false
     lateinit var sizeConstraint: SizeConstraint
     lateinit var durationConstraint: DurationConstraint
+    var isNeedGeo = false
```
AlexV525, line 33: Consist variables name everywhere.

android/src/main/kotlin/com/fluttercandies/photo_manager/core/utils/AndroidQDBUtils.kt
```diff
@@ -248,21 +254,32 @@ object AndroidQDBUtils : IDBUtils {
...
+                if (getGeo) {
+                    it.latLong?.apply {
+                        lat = this[0]
+                        log = this[1]
                     }
+                    getGeo = false
+                }
+            }
+        }
+        if (getGeo) {
```
AlexV525, line 277: The EXIF interface will be created again here. The situation is even worse.

lib/src/filter/filter_options.dart
```diff
@@ -13,6 +13,7 @@ import '../internal/enums.dart';
 class FilterOption {
   const FilterOption({
     this.needTitle = false,
+    this.needGeo = false,
```
AlexV525, line 16: Better be `needLatLng`.

lib/src/internal/plugin.dart
```diff
@@ -123,6 +123,7 @@ class PhotoManagerPlugin with BasePlugin, IosPlugin, AndroidPlugin {
     required int start,
     required int end,
     required FilterOptionGroup optionGroup,
+    bool forceGetLocation = false,
```
AlexV525, line 126: Remove useless code.

AlexV525 (changes requested):

AlexV525:
The solution is not in a good state IMO. Please file a new PR after the design is more solid and clear. Thanks.

PR #799, opened 2022-07-13
fix: 🐛 startChangeNotify don't work on iOS Device

bay2:
image
There is no way to keep going , because on Objc code "notify" method never return result.


AlexV525:
This needs to fix on both sides. I'll submit a new PR for this. Thanks for the heads up.

PR #839, opened 2022-09-30
fix: handle detach from engine for iOS

AlessandroToschi:


AlexV525:
Hi, any reason that you abandon the pull request?

AlessandroToschi:
Sorry, I opened the PR by mistake.

My idea was first to open an issue reporting the crash we're experiencing and then provide a PR if welcomed.

AlessandroToschi:
The main idea is to propagate the `detachFromEngine` to any entity that creates a method channel and then set to nil the method handler when the engine is torn down.

ios/Classes/PMNotificationManager.m
```diff
@@ -42,6 +42,13 @@ - (void)stopNotify {
     _notifying = NO;
 }
 
+-(void)detachFromEngine {
+    if (_notifying) {
+        [self stopNotify];
```
AlexV525, line 47: I'm not sure whether we should stop notifying here.

ios/Classes/PMPlugin.m
```diff
@@ -16,22 +16,38 @@
 @implementation PMPlugin {
     BOOL ignoreCheckPermission;
     NSObject <FlutterPluginRegistrar> *privateRegistrar;
+    FlutterMethodChannel* channel;
+    NSMutableArray* progressHandlerDisposeCallbacks;
```
AlexV525, line 20: Can we have a better structure here? Setting up an array for fallbacks seems overhead for the plugin instance.

ios/Classes/PhotoManagerPlugin.m
```diff
@@ -3,9 +3,20 @@
...
 }
+
+@synthesize plugin;
+@synthesize registrar;
+
 + (void)registerWithRegistrar:(NSObject <FlutterPluginRegistrar> *)registrar {
-    PMPlugin *plugin = [PMPlugin new];
-    [plugin registerPlugin:registrar];
+    PhotoManagerPlugin* photoManagerPlugin = [[PhotoManagerPlugin alloc] init];
+    photoManagerPlugin.plugin =  [[PMPlugin alloc] init];
```
AlexV525, line 12:
```suggestion
    photoManagerPlugin.plugin = [[PMPlugin alloc] init];
```

ios/Classes/PMNotificationManager.m
AlessandroToschi: Well, actually the `FlutterPlugin detachFromEngine` is the cleanup procedure during the engine deallocation, which is usually an action performed during the exit of the app. So leaving the notification on but the method channel is not available I don't think is usable in any way since that method channel will never be available again.

ios/Classes/PMPlugin.m
AlessandroToschi: What about the classic delegate pattern? 
Creating a delegate that `PMPlugin` will implement and then passing it to the progress handler?

CaiJingLong:
I try to run the example.
The `weakSelf` is always nil.
image

CaiJingLong (changes requested):
I think the code for weak references needs to be fixed.

AlexV525:
Closing since we didn't get valid feedback for a while.

PR #863, opened 2022-11-09
⚡️ Improve media subtype on iOS

AlexV525:


v1ktorstep:
I think something wrong with subtypes now. Video subtype values dosn't correspond native values. For example for video I should get value 131072, but I get 1048576.

Screen Shot 2022-12-29 at 16 36 46

AlexV525:
> For example for video with subtype PHAssetMediaSubtypeVideoHighFrameRate I should get value 131072, but I get 1048576.

You should not depend on specific subtype values. And it's now directly returned from the native definition.

v1ktorstep:
> You should not depend on specific subtype values.

How can I use this then? For example, if I want display only screenshots (on ios) I can check if subtype == 1 << 2, and it's work fine, but not for video subtypes.
Screen Shot 2022-12-29 at 18 45 07

PR #913, opened 2023-03-25
:memo: Update doc

CaiJingLong:


CaiJingLong:
The PR contains a check_license.dart.
After this PR is passed, I will add a workflow to use it to detect whether the dart file in lib contains license file header.

PR #944, opened 2023-07-13
fix: Support only add permission for iOS

CaiJingLong:


example/ios/Runner/Info.plist
```diff
@@ -14,8 +14,6 @@
 	<string>6.0</string>
 	<key>CFBundleName</key>
 	<string>com.fluttercandies.photo_manager_example</string>
-	<key>CFBundleDisplayName</key>
-	<string>photo_manager Example</string>
```
AlexV525, line 18: Why is this being deleting?

CaiJingLong: image

I don't seem to have manually removed it.
I guest the xcode delete it when compile, because it repeats.

AlexV525 (approved):

PR #945, opened 2023-07-17
Fix: the problem with duplicate filename of iOS

CaiJingLong:


CaiJingLong:
The new file name contains id of PHAsset.

image

ios/Classes/core/PMManager.m
```diff
@@ -479,6 +479,12 @@ - (void)getFullSizeFileWithId:(NSString *)id isOrigin:(BOOL)isOrigin subtype:(in
                 return;
             }
         }
+        if (@available(macOS 14.0, *)) {
+            if (asset.isLivePhoto && (subtype & PHAssetMediaSubtypePhotoLive) == PHAssetMediaSubtypePhotoLive) {
+                [self fetchLivePhotosFile:asset handler:handler progressHandler:progressHandler];
+                return;
+            }
+        }
```
CaiJingLong, line 487: When I was looking at the problem with the file name, I found that macOS and iOS have different logic about getting livePhoto, so I add the code.

CaiJingLong:
In Android, the access permission itself depends on the Activity, so the code contains the code for dynamic permission checking, which requires your FlutterView to be Attached to an Activity, and it is performed according to the standard Android specification (that is, the channel called by each dart methods are wrapped in code for dynamic permission checking).

And your background service does not seem to depend on Activity, so you need to use ignorePermissionCheck to ignore PhotoManager's permission check on the premise of ensuring that you have permission.

```dart
// Call the method to ignore permission check
await PhotoManager.setIgnorePermissionCheck(true);
```

ios/Classes/core/PMManager.m
```diff
@@ -714,7 +721,7 @@ - (NSString *)makeAssetOutputPath:(PHAsset *)asset
     } else {
         filename = [asset valueForKey:@"filename"];
     }
-    filename = [NSString stringWithFormat:@"%@%@_%@", modifiedDate, isOrigin ? @"_o" : @"", filename];
+    filename = [NSString stringWithFormat:@"%@_%@%@_%@", id, modifiedDate, isOrigin ? @"_o" : @"", filename];
```
AlexV525, line 724: Would it be too long if we added the ID here?

CaiJingLong: Other suggestion?

The id of asset will add 44 chars.
The biggest of darwin filename is 255.

AlexV525: Maybe compress the ID into 6~8 characters ID?

CaiJingLong: I'm not sure what is the probability that a certain substring of the id will be repeated.

AlexV525 (approved):

PR #958, opened 2023-07-29
Update AndroidManifest.xml

projekt01:
Removed package attribute. It's a part of the gradle.build file now.
See: https://developer.android.com/build/publish-library/prep-lib-release


AlexV525:
What's the compatibility of removing this? More specifically, what is the lowest version of AGP that can recognize this?

AlexV525 (approved):

PR #959, opened 2023-07-29
fix: delete assets permanently deleted instead of trash in android R

shalong-tanwen:
### Changes made

The current implementation of trashing the asset instead of deleting it in Android 30 and above feels inconsistent with the rest of the behavior of the API in older versions of Android and iOS where the asset is irrevocably deleted.

It is changed in this PR to have the same behavior of deleting the asset using `MediaStore.createDeleteRequest` which based on the [documentation](https://developer.android.com/reference/android/provider/MediaStore#createDeleteRequest(android.content.ContentResolver,%20java.util.Collection%3Candroid.net.Uri%3E)) calls [ContentResolver#delete](https://developer.android.com/reference/android/content/ContentResolver#delete(android.net.Uri,%20android.os.Bundle)) similar to the [handling](https://github.com/fluttercandies/flutter_photo_manager/blob/ba688606fddaa910b1ee92b85ff7159c60ebf983/android/src/main/kotlin/com/fluttercandies/photo_manager/core/PhotoManagerDeleteManager.kt#L48) used in older android versions

### Possible improvements
A new method named `trashAsset` can be added as an feature which would trash the asset instead of deleting it in Android 30 and above along with a method to fetch such assets as well

### DIFF
--- a/android/src/main/kotlin/com/fluttercandies/photo_manager/core/PhotoManagerDeleteManager.kt
+++ b/android/src/main/kotlin/com/fluttercandies/photo_manager/core/PhotoManagerDeleteManager.kt
@@ -57,7 +57,7 @@ class PhotoManagerDeleteManager(val context: Context, private var activity: Acti
     @RequiresApi(Build.VERSION_CODES.R)
     fun deleteInApi30(uris: List<Uri?>, resultHandler: ResultHandler) {
         this.androidRHandler = resultHandler
-        val pendingIntent = MediaStore.createTrashRequest(cr, uris.mapNotNull { it }, true)
+        val pendingIntent = MediaStore.createDeleteRequest(cr, uris.mapNotNull { it })
         activity?.startIntentSenderForResult(
             pendingIntent.intentSender,
             androidRDeleteRequestCode,


shalong-tanwen:
Fixed the Android CI failing due to the lack of namespace in the example app's `build.gradle` as well

CaiJingLong:
I need to observe the real performance on the virtual machine, I will test this modification in the near future.


---

> A new method named trashAsset can be added as an feature which would trash the asset instead of deleting it in Android 30 and above along with a method to fetch such assets as well

Introducing new dart api is really a good idea.

example/android/app/build.gradle
```diff
@@ -32,6 +32,8 @@ apply plugin: 'kotlin-kapt'
 apply from: "$flutterRoot/packages/flutter_tools/gradle/flutter.gradle"
 
 android {
+    namespace 'com.flutterandies.photo_manager_example'
```
CaiJingLong, line 35: Code here needs to resolve conflicts because of #961.

CaiJingLong (changes requested):

example/android/app/build.gradle
shalong-tanwen: Conflict resolved.

CaiJingLong (approved):

PR #969, opened 2023-08-09
feature-filesize

droplet-js:
add filesize support


CaiJingLong:
For performance reasons, this PR is unlikely to pass.
@AlexV525

AlexV525:
Providing the file size along with entities would cause huge performance regressions during list methods. We already have various files getter so it would be easy to get the file size from them. I haven't tested it but as I imagine this might cause infinite waiting if some entities are located in the iCloud only and cannot be fetched smoothly.

AlexV525:
Closing based on the above comment.

PR #975, opened 2023-08-21
Try catch request cache thumbnail

masewo:
App crashes when photo_manager tries to request a cache thumbnail for a broken image file, like this one:
https://drive.google.com/file/d/1-8C4w2fhbHh2EmsJPF2SN0de3NiTNiCN/view?usp=drivesdk

Catch the thrown exception so that the main app does not crash.


CaiJingLong (approved):
LGTM

PR #991, opened 2023-09-21
Update for ImageProvider deprecation

moffatman:
Now since flutter 3.15, we can't just override `ImageProvider.load`, and `DecoderCallback` is removed. The deprecation occured in flutter 2.13. So would have to also upgrade minimum flutter sdk version in pubspec.yaml I guess?

### DIFF
--- a/lib/src/internal/image_provider.dart
+++ b/lib/src/internal/image_provider.dart
@@ -66,9 +66,9 @@ class AssetEntityImageProvider extends ImageProvider<AssetEntityImageProvider> {
   ImageFileType get imageFileType => _getType();
 
   @override
-  ImageStreamCompleter load(
+  ImageStreamCompleter loadImage(
     AssetEntityImageProvider key,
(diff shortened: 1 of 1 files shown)


AlexV525:
Could you rebase to tip of tree?

AlexV525:
Also, we typically dont support versions other than the stable branch.

CaiJingLong:
Although we plan to remove the ImageProvider in the package in version 3.0, can we provide a new package before then?
@AlexV525

AlexV525:
> Although we plan to remove the ImageProvider in the package in version 3.0, can we provide a new package before then? @AlexV525

Sounds a great idea.

AlexV525:
So we're working on separating the image provider as an external package. Thus we no longer accept changes related to it. Thanks for the contributions. @moffatman @shkvoretz

moffatman:
Sure, I don't use it, just published this PR to fix compilation of the package

PR #993, opened 2023-09-24
✨ Add locks to the image provider

AlexV525:
Resolves https://github.com/LeGoffMael/insta_assets_picker/issues/24.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ that can be found in the LICENSE file. -->
 
 # CHANGELOG
 
+## 2.7.2
+
+### Improvements
+
+- Add locks to the image provider.
+
 ## 2.7.1
 
 ### Fixes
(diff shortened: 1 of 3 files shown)


AlexV525:
@LeGoffMael Could you help test the fix?

CaiJingLong:
This is an attempt to solve the concurrency problem of ImageProvider from the dart layer.
Android uses Glide to load thumbnails, and there are no network images, so there are no concurrency issues

iOS has icloud and PHAssetResource and caches them locally, so there will be some problems.
I plan to build a simple local caching natively on iOS to solve this problem in the future. Before that, we can try to use the code of this PR.

LeGoffMael:
Hi @AlexV525, i will check it later tonight

LeGoffMael:
I just tested using `insta_assets_picker` and it seems that the error is fixed !

I tried to load assets on my iPhone (from local and iCloud) multiple times and i could not reproduce the problem.
I restarted m phone, clean the project and delete the example app between each test (with and without the fix).

PR #1003, opened 2023-10-11
Create multiple albums with the same name

liugaocomeon:
When creating an album, there may be situations where an album with the same name is created. Before creating, you should check whether an album with the same name exists.


CaiJingLong:
If you want to solve the problem of the same name, you should consider all (including user-created), not just smart albums.

CaiJingLong (changes requested):
If it is a duplicate name problem, it should include smart photo albums and user photo albums.

See [doc](https://developer.apple.com/documentation/photokit/phassetcollectiontype?language=objc) of apple.

PR #982, opened 2023-09-01
Android-14-permission

CaiJingLong:

PR #1004, opened 2023-10-13
feat!(darwin): Remove permission check code

CaiJingLong:


CaiJingLong:
Removed runtime permission checking code to be consistent with #982.

PR #1055, opened 2023-12-08
fix: auth value for android API 34

CaiJingLong:
Signed-off-by: CaiJingLong <cjl_spy@163.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ that can be found in the LICENSE file. -->
 
 # CHANGELOG
 
+## 3.0.0-dev.5
+
+- Fix: the requestPermissionExtend method returns the wrong result on android API 34.
+
 ## 3.0.0-dev.4
 
 - Fix presentLimit have no result on android API 34. [#1051](https://github.com/fluttercandies/flutter_photo_manager/issues/1051)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,9 @@ To know more about breaking changes, see the [Migration Guide][].
 
 ## 3.0.0-dev.5
 
-- Fix: the requestPermissionExtend method returns the wrong result on android API 34.
+### Fixes
+
+- Fix `requestPermissionExtend` returns the incorrect status on Android API 34.
 
 ## 3.0.0-dev.4
 
--- a/android/src/main/kotlin/com/fluttercandies/photo_manager/permission/impl/PermissionDelegate34.kt
+++ b/android/src/main/kotlin/com/fluttercandies/photo_manager/permission/impl/PermissionDelegate34.kt
@@ -32,6 +32,11 @@ class PermissionDelegate34 : PermissionDelegate() {
         requestType: Int,
         mediaLocation: Boolean
     ) {
+        if (havePermissions(context, requestType)) {
+            permissionsUtils.permissionsListener?.onGranted(mutableListOf())
+            return
+        }
+
         LogUtils.info("requestPermission")
         var havePermission = true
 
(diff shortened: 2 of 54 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ that can be found in the LICENSE file. -->
 
 To know more about breaking changes, see the [Migration Guide][].
 
+## 3.0.0-dev.5
+
+- Fix: the requestPermissionExtend method returns the wrong result on android API 34.
```
AlexV525, line 11:
```suggestion
### Fixes

- Fix `requestPermissionExtend` returns the incorrect status on Android API 34.
```

android/src/main/kotlin/com/fluttercandies/photo_manager/permission/impl/PermissionDelegate34.kt
```diff
@@ -32,50 +32,55 @@ class PermissionDelegate34 : PermissionDelegate() {
...
+
+        LogUtils.info("requestPermission")
         var havePermission = true
 
         val containsImage = RequestTypeUtils.containsImage(requestType)
         val containsVideo = RequestTypeUtils.containsVideo(requestType)
         val containsAudio = RequestTypeUtils.containsAudio(requestType)
 
-        val permissions = mutableListOf<String>()
+        val needPermissions = mutableListOf<String>()
```
AlexV525, line 47:
```suggestion
        val requiredPermissions = mutableListOf<String>()
```

lib/src/types/types.dart
```diff
@@ -60,6 +60,20 @@ class RequestType {
...
+    audio,
+  ];
+
+  static RequestType fromTypes(List<RequestType> types) {
+    var result = const RequestType(0);
+    for (final type in types) {
+      result += type;
+    }
+    return result;
+  }
```
AlexV525, line 75:
```suggestion
  static const values = <RequestType>[image, video, audio];

  /// Computes the request type from given types.
  static RequestType fromTypes(List<RequestType> types) {
    RequestType result = const RequestType(0);
    for (final type in types) {
      result += type;
    }
    return result;
  }
```

pubspec.yaml
```diff
@@ -12,7 +12,7 @@ dependencies:
     sdk: flutter
 
 dev_dependencies:
-  flutter_lints: any
+  flutter_lints: ^3.0.1
```
AlexV525, line 15: We can still use `any`?

CaiJingLong: Because the project did not upload the pubspec.lock file, the local usage deps and the remote CI were no same, so CI often failed to pass of the lints.
I recommend using the specified version, does it have any other effects?

CHANGELOG.md
CaiJingLong: The same tag will affect the generation of markdown toc and trigger markdown lints. I personally do not recommend using multiple identical secondary headers.

pubspec.yaml
AlexV525: Because the package resolves different Dart SDK between its major versions:
- 3.x: >=3.0.0
- 2.x: >=2.17.0
- 1.x: >=2.12.0

Since our package supports `>=2.13.0`, using `any` will help to develop the package with different SDK versions.

CHANGELOG.md
CaiJingLong: Although, I turned off this lint, the repeated headers have a negative effect on toc generation.

https://github.com/DavidAnson/markdownlint/blob/v0.32.1/doc/md024.md

AlexV525: Well, we don't need a TOC for the CHANGELOG. :D

pubspec.yaml
CaiJingLong: As far as I know, reference relationships of dev_dependencies are not transitive.
https://dart.dev/tools/pub/dependencies#dev-dependencies

CaiJingLong: However, changing it back is not a big problem. If you encounter CI problems in the future, just run `dart fix`.

AlexV525 (approved):

PR #1114, opened 2024-04-08
feat: Upgrade android configs

CaiJingLong:
- Upgrade AGP version to 8.3.1
- Upgrade kotlin version to 1.9.23
- Upgrade gradle wrapper version to 8.4
- Use `kotlin.jvmToolchain` to set android.compileOptions.

The `jvmToolchain` document:  https://developer.android.com/build/jdks?hl=zh-cn


android/settings.gradle
```diff
@@ -1 +1,7 @@
 rootProject.name = 'photo_manager'
+
+pluginManagement {
+    repositories {
+        gradlePluginPortal()
+    }
+}
```
AlexV525, line 7:
```suggestion
}

```

PR #1149, opened 2024-06-21
[iOS] Handle error when fetching thumbnail

maksg:
One of my videos can't generate a thumbnail and it causes the error `Domain=PHPhotosErrorDomain Code=3303`, resulting in `fetchThumb` never calling back.
We should handle this case probably the same way as when `imageData` is `nil`.

I'm attaching the video I have issue with.

https://github.com/fluttercandies/flutter_photo_manager/assets/11507974/b2d3d444-072a-4e50-8060-ef61e172fe88

### DIFF
--- a/ios/Classes/core/PMManager.m
+++ b/ios/Classes/core/PMManager.m
@@ -440,6 +440,12 @@
                       contentMode:option.contentMode
                           options:requestOptions
                     resultHandler:^(PMImage *result, NSDictionary *info) {
+        if (info[PHImageErrorKey]) {
+            [handler reply: nil];
+            [self notifySuccess:progressHandler];
+            return;
+        }
+
         BOOL downloadFinished = [PMManager isDownloadFinish:info];
 
         if (!downloadFinished) {


AlexV525:
Your video only lasts 2 seconds and seems unrelated.

ios/Classes/core/PMManager.m
```diff
@@ -440,15 +440,22 @@ - (void)fetchThumb:(PHAsset *)asset option:(PMThumbLoadOption *)option resultHan
                       contentMode:option.contentMode
                           options:requestOptions
                     resultHandler:^(PMImage *result, NSDictionary *info) {
-        BOOL downloadFinished = [PMManager isDownloadFinish:info];
+        if ([handler isReplied]) {
+            return;
+        }
```
AlexV525, line 445: This is dead code, no?

ios/Classes/core/PMManager.m
```diff
@@ -440,15 +440,22 @@ - (void)fetchThumb:(PHAsset *)asset option:(PMThumbLoadOption *)option resultHan
...
 
-        if (!downloadFinished) {
+        if (info[PHImageErrorKey]) {
+            [handler reply: nil];
+            [self notifySuccess:progressHandler];
```
AlexV525, line 449: Should be `[self notifyProgress:progressHandler progress:progress state:PMProgressStateFailed];` ?

maksg:
This is the video that I can't fetch thumbnail from. Add it to your library and call `fetchThumb` on it, you should get an error and no callback.

ios/Classes/core/PMManager.m
```diff
@@ -440,15 +440,22 @@ - (void)fetchThumb:(PHAsset *)asset option:(PMThumbLoadOption *)option resultHan
...
             return;
         }
 
-        if ([handler isReplied]) {
+        BOOL downloadFinished = [PMManager isDownloadFinish:info];
+
+        if (!downloadFinished) {
```
AlexV525, line 455: Can we inline this?

ios/Classes/core/PMManager.m
maksg: I have no idea. I didn't want to refactor your code too much. I just moved it to the top since I didn't want `[handler reply: nil];` to be called more than once. But I can remove that if you think it's a dead code.

ios/Classes/core/PMManager.m
maksg: Do you mean like `if (![PMManager isDownloadFinish:info]) {`? I can do that but it will be inconsistent with other parts of your code.

ios/Classes/core/PMManager.m
AlexV525: I took a look at the previous, so you basically just changing the order of the flow and adding this predicate. Can you simply add this predicate at the top of the callback?

maksg: Sure thing, won't that cause any issues if it gets called twice or are we sure it won't get called twice? I changed the order because `if ([handler isReplied])` was preventing multiple calls to `[handler reply: data];` earlier and I wanted to be safe when calling this myself.

AlexV525: Ah cool. I like the idea!

maksg: Ah, I saw just a like and thought it meant that I should change that anyway. But I reverted it now. In this case is there anything you want me to change still or is this fine?

AlexV525: Yes please change it. Nothing should be changed if the tests are happy.

maksg: Looks like some Android tests are not passing but it doesn't seem to be related to my changes.

PR #1184, opened 2024-09-13
feat: get current permission

CaiJingLong:


ios/Classes/PMPlugin.m
```diff
@@ -121,9 +129,27 @@ - (void)handleNotNeedPermissionMethod:(ResultHandler *)handler {
...
         [manager clearCache];
         [handler reply:nil];
+    } else if ([method isEqualToString:@"getPermissionState"]) {
+        [self getPermissionState:handler];
     }
 }
 
+- (void)getPermissionState:(ResultHandler *)handler {
+    int requestAccessLevel = [handler.call.arguments[@"iosAccessLevel"] intValue];
+#if __IPHONE_14_0
```
AlexV525, line 139: Is the macro required? The minimum Xcode has already supported at least iOS 17.

lib/src/internal/enums.dart
```diff
@@ -63,7 +63,7 @@ enum OrderOptionType {
 enum PMRequestState { prepare, loading, success, failed }
 
 /// Information about app’s authorization to access the user’s photo library.
-///  * Android: Only [authorized] and [denied] are valid.
+///  * Android: Have [authorized], [denied] and [limited]
```
AlexV525, line 66:
```suggestion
///  * Android: [authorized], [denied], and [limited].
```

lib/src/internal/plugin.dart
```diff
@@ -665,6 +665,16 @@ class PhotoManagerPlugin with BasePlugin, IosPlugin, AndroidPlugin, OhosPlugin {
...
   }
+
+  Future<PermissionState> getPermissionState(
+    PermissionRequestOption requestOption,
+  ) async {
+    final int? result = await _channel.invokeMethod<int>(
+      PMConstants.mGetPermissionState,
+      requestOption.toMap(),
+    );
+    return PermissionState.values[result ?? 0];
```
AlexV525, line 676: Is there any chance that we can get a nullable result here? If not, use the non-nullable return.

lib/src/managers/photo_manager.dart
```diff
@@ -57,6 +57,35 @@ class PhotoManager {
...
+  /// PermissionState state = await PhotoManager.getPermissionState();
+  /// if (state == PermissionState.authorized) {
+  ///   print('The application has full access permission');
+  /// } else {
+  ///   print('The application does not have full access permission');
+  /// }
+  /// ```
+  ///
+  /// Note: On Android, this method may require an [Activity] context. If called in a background service,
+  /// make sure to call the [setIgnorePermissionCheck] method first.
```
AlexV525, line 82:
```suggestion
  /// Get the current [PermissionState] of the photo library.
  ///
  /// Example:
  /// ```dart
  /// PermissionState state = await PhotoManager.getPermissionState();
  /// if (state == PermissionState.authorized) {
  ///   print('The application has full access permission');
  /// } else {
  ///   print('The application does not have full access permission');
  /// }
  /// ```
  ///
  /// Note: On Android, this method may require an [Activity] context.
  /// Call [setIgnorePermissionCheck] first if the call is from a background service.
```

lib/src/managers/photo_manager.dart
```diff
@@ -57,6 +57,35 @@ class PhotoManager {
...
+  /// Get the current permission state of the application
+  ///
+  /// This method returns a [Future] that resolves to the current [PermissionState] of the application.
+  ///
+  /// Possible values for [PermissionState] include:
+  /// - [PermissionState.notDetermined]: The user hasn't made a choice yet
+  /// - [PermissionState.restricted]: The application is restricted from accessing the photo library (iOS only)
+  /// - [PermissionState.denied]: The user has denied permission
+  /// - [PermissionState.authorized]: The user has granted full access permission
+  /// - [PermissionState.limited]: The user has granted limited access permission (iOS 14+ only)
```
AlexV525, line 69: We could state these values in the enum definition.

lib/src/managers/photo_manager.dart
```diff
@@ -57,6 +57,35 @@ class PhotoManager {
...
+  /// } else {
+  ///   print('The application does not have full access permission');
+  /// }
+  /// ```
+  ///
+  /// Note: On Android, this method may require an [Activity] context. If called in a background service,
+  /// make sure to call the [setIgnorePermissionCheck] method first.
+  static Future<PermissionState> getPermissionState({
+    PermissionRequestOption requestOption = const PermissionRequestOption(),
+  }) {
```
AlexV525, line 85: Could we make `requestOption` as a required named parameter? In that way we let developers always have a sense of what they are trying to do and provide valid input.

lib/src/types/types.dart
```diff
@@ -82,7 +82,20 @@ class RequestType {
...
+      values.add('image');
+    }
+    if (containsVideo()) {
+      values.add('video');
+    }
+    if (containsAudio()) {
+      values.add('audio');
+    }
+
+    return '$runtimeType(${values.join(', ')})';
```
AlexV525, line 97:
```suggestion
    return 'RequestType(${values.join(', ')})';
```

PR #1188, opened 2024-09-18
Fix MP4 videos saving result in 3GP on Android API <=28

yishangfei:
Fixed the issue that the video format changes to 3gp when saving mp4 videos to the album on Android 9


AlexV525:
Could you provide a test case in which saving 3gp will fail? Your changes don't tell the idea and why it will fix the issue well.

yishangfei:
### pubspec.yaml
photo_manager: ^3.3.0
photo_manager_image_provider: ^2.1.1

image
image

### use my code  
e0bd912a05e01fcca51f557d9eb0d26

In MI 8 Lite android 9

AlexV525:
It is still unclear why the PR fixes the issue. Could you provide a minimal reproducible example?

yishangfei:
var appCacheDir = await getApplicationCacheDirectory();
          String filePath = '${appCacheDir.path}/aaa.mp4';
          File result = File(filePath);
          var entity = await PhotoManager.editor.saveVideo(
            result,
            title: path.basename(result.path),
          );
          print('assetEntity===${entity!.title}===${await entity.file}');

yishangfei:
### example code
```
var appCacheDir = await getApplicationCacheDirectory();
String filePath = '${appCacheDir.path}/aaa.mp4';
File result = File(filePath);
var entity = await PhotoManager.editor.saveVideo(
      result,
     title: path.basename(result.path),
     );  
print('assetEntity===${entity!.title}===${await entity.file}');
```
On Android 9, the source file is mp4 and converted to 3gp format through saveVideo

AlexV525:
Is this happening on all videos or just one specific video? Could you file an issue to summarize all the details? And provide a minimal runnable example so we can run to reproduce.

yishangfei:
All videos on android9

demo: https://github.com/yishangfei/test1 

image

android/src/main/kotlin/com/fluttercandies/photo_manager/core/utils/IDBUtils.kt
```diff
@@ -428,11 +429,6 @@ interface IDBUtils {
         }
         refreshStream()
 
-        val shouldKeepPath = if (!isAboveAndroidQ) {
```
AlexV525, line 431: Could you also keep the condition? The fix should work without removing it, correct?

android/src/main/kotlin/com/fluttercandies/photo_manager/core/utils/IDBUtils.kt
```diff
@@ -493,6 +490,37 @@ interface IDBUtils {
         return getAssetEntity(context, id.toString())
     }
 
+    private fun getAlbumFolderPath(
+        folderName: String?,
+    ): String {
+        var albumFolderPath: String = Environment.getExternalStorageDirectory().path
+        if (android.os.Build.VERSION.SDK_INT < 29) {
```
AlexV525, line 497: What standard do we following here? Should we use `MOVIE` as much as possible?

android/src/main/kotlin/com/fluttercandies/photo_manager/core/utils/IDBUtils.kt
```diff
@@ -493,6 +490,37 @@ interface IDBUtils {
...
+            createDirIfNotExist(
+                Environment.getExternalStoragePublicDirectory(baseFolderName).path
+            ) ?: albumFolderPath
+        } else {
+            createDirIfNotExist(albumFolderPath + File.separator + folderName) ?: albumFolderPath
+        }
+        return albumFolderPath
+    }
+
+    private fun createDirIfNotExist(dirPath: String): String? {
```
AlexV525, line 511: We seem to already have a helper method `checkDir`

yishangfei:
I added it in line 455 according to your latest code.

android/src/main/kotlin/com/fluttercandies/photo_manager/core/utils/IDBUtils.kt
```diff
@@ -454,6 +452,14 @@ interface IDBUtils {
                 if (relativePath.isNotBlank()) {
                     put(RELATIVE_PATH, relativePath)
                 }
+            }else{
+                val albumDir = "${Environment.getExternalStorageDirectory().path}${File.separator}${Environment.DIRECTORY_MOVIES}"
+                // check if the directories exist
+                "$albumDir${File.separator}$title".checkDirs()
+                val timestamp = System.currentTimeMillis().toString()
+                // Create a video file. If you use file.name, assetEntity may be empty.
```
AlexV525, line 460: Why we cannot use the filename?

AlexV525:
Still working on this?

yishangfei:
now there is no problem. I only modified the version below Android 10. It’s just that the timestamp needs to be used when saving. If the original name is used a second time, the entity returned by PhotoManager.editor.saveVideo will be null.

AlexV525:
> If the original name is used a second time, the entity returned by PhotoManager.editor.saveVideo will be null.

We need to figure out why.

Also please do another rebase on the latest changes and remove unnecessary changes.

yishangfei:
The automatic renaming function (for example, video (1).mp4) is implemented in Android 11 and above. In earlier versions, if you repeat cr.insert in the insertUri method, an exception will be thrown. "Cannot insert the new asset."

```
 val uri = cr.insert(contentUri, values) ?: throw RuntimeException("Cannot insert the new asset.")
```

Just like the pictures I provided before，versions below Android Q use timestamp naming to insert, or write a renaming logic that is the same as that of higher versions.

AlexV525:
> The automatic renaming function (for example, video (1).mp4) is implemented in Android 11 and above.

TIL. Are there any docs referencing this?

yishangfei:
I have not been able to find any official documentation.
I found the solution on stackoverflow.
[ContentResolver.insert always returning null](https://stackoverflow.com/questions/22419867/contentresolver-insert-always-returning-null)

When using the ContentResolver.insert() method to insert a file with the same name, an exception will indeed be thrown if the file with the same name already exists. This is because in Android 10 and below, MediaStore does not have the automatic renaming function.

AlexV525:
> add the following comment to the code. Is it OK?

How about:

```
Using a duplicate filename that already exists on the device will cause the insertion to fail on Android API 30-.
```

yishangfei:
> > add the following comment to the code. Is it OK?
> 
> How about:
> 
> ```
> Using a duplicate filename that already exists on the device will cause the insertion to fail on Android API 30-.
> ```

i submitted

AlexV525:
I've also added some format changes and fixed the nullable saving result.

AlexV525 (approved):
Thanks for your contribution.

PR #1204, opened 2024-10-15
🐛 Fix no returned ids after successful deletion on Android API 29+

chocoyang-horizon:
When the file is created by the application itself, the uri can be deleted successfully without causing a RecoverableSecurityException.

[Note]: I tried Android 13 and the same way can also delete files created by the application without triggering the dialog. Maybe developers can decide in the future whether to force the use of createDeleteRequest method. (optional)


AlexV525 (approved):
Thanks for the fix

PR #1205, opened 2024-10-15
🐛 Fix mediaLocation = true does not work on Android API 34

chocoyang-horizon:
mediaLocation = true does not work when the app already has common permissions on Android API 34


AlexV525:
Does "not work" indicate that permission for the media location won't be requested in this case?

AlexV525 (approved):

chocoyang-horizon:
How to reproduce:

The app has not used the mediaLocation permission in the past.
```dart
final state = await PhotoManager.requestPermissionExtend();
```

Now the app requires the mediaLocation permission, but the plugin doesn't actually request the mediaLocation permission dynamically.
```dart
final state = await PhotoManager.requestPermissionExtend(
  requestOption: const PermissionRequestOption(
    androidPermission: AndroidPermission(
      type: RequestType.common,
      mediaLocation: true,
    ),
  ),
);
```

PR #1229, opened 2024-11-23
Fix the issue where album assets cannot be retrieved when selecting only and video or only and image conditions

RW199908:
Fix the issue where album assets cannot be retrieved when selecting only and video or only and image
https://github.com/fluttercandies/flutter_photo_manager/issues/1230#issue-2685220253


AlexV525:
Could you provide screen recordings with the issue happen and after applied the fix?

RW199908:
Uploading meeting_1.mp4…

delfme:
Hi @RW199908 , couldn't check the code, would this be helful with https://github.com/fluttercandies/flutter_photo_manager/issues/1257 ?

PR #1259, opened 2025-01-21
FEAT: change SizeConstraint ignoreSize defualt to true

6a209:
The expression "0 < width < 100000" has the same effect as "ignoreSize = true" but adding these two conditions in SQL might degrade performance somewhat.

### DIFF
--- a/lib/src/filter/classical/filter_options.dart
+++ b/lib/src/filter/classical/filter_options.dart
@@ -109,7 +109,7 @@ class SizeConstraint {
     this.maxWidth = 100000,
     this.minHeight = 0,
     this.maxHeight = 100000,
-    this.ignoreSize = false,
+    this.ignoreSize = true,
   });
 
   /// The minimum width for queried images.


AlexV525:
Changing the flag would be a breaking change for users who wrote a valid condition but didn't toggled the flag before.

6a209:
> Changing the flag would be a breaking change for users who wrote a valid condition but didn't toggled the flag before.
But it originally felt like there was an error; the default value theory should be unconditional, which would be more reasonable. However, a condition (0 < width < 100000) was added by default. If I really have an image larger than 100000, wouldn't it fail to be filtered out? I think this is not the expected behavior of the user. But then again, 100000 is really too large, almost non-existent such big images, and if such an image were to be loaded into memory all at once, it would definitely cause outofmemory,

AlexV525:
> But it originally felt like there was wrong, the default value theory should be unconditional, which would be more reasonable. 

If we could convert `100000` to `null`, that would be more reasonable.

PR #1288, opened 2025-05-14
feat: Add includeHiddenAssets option (新增包含隐藏资源选项)

CaiJingLong:
- Add `includeHiddenAssets` property to `FilterOptionGroup` and `PMFilter` classes. (在 `FilterOptionGroup` 和 `PMFilter` 类中添加 `includeHiddenAssets` 属性。)
- Add `includeHiddenAssets` checkbox to `FilterOptionPage` to control the visibility of hidden assets on iOS. (在 `FilterOptionPage` 中添加 `includeHiddenAssets` 复选框，用于控制 iOS 上隐藏资源的可见性。)
- Create `IncludeHiddenTestPage` to test the count of hidden and non-hidden assets. (创建 `IncludeHiddenTestPage` 页面，用于测试 隐藏和非隐藏资源的计数。)
- Modify `PMConvertUtils.m` and `PMFilterOption.m` to handle the `includeHiddenAssets` option on iOS. (修改 `PMConvertUtils.m` 和 `PMFilterOption.m` 以处理 iOS 上的 `includeHiddenAssets` 选项。)
- Update example app to include the new test page in `CustomFilterExamplePage`. (更新示例应用程序，在 `CustomFilterExamplePage` 中包含新的测试页面。)

### DIFF
--- a/lib/src/filter/base_filter.dart
+++ b/lib/src/filter/base_filter.dart
@@ -60,9 +60,9 @@ abstract class PMFilter {
   bool containsPathModified = false;
 
   /// Whether to include hidden assets in the results.
-  /// 
+  ///
   /// This option only takes effect on iOS.
-  /// 
+  ///
   /// See also:
   ///  * [PHFetchOptions.includeHiddenAssets](https://developer.apple.com/documentation/photos/phfetchoptions/includehiddenassets).
   bool includeHiddenAssets = false;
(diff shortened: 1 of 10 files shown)


lib/src/filter/base_filter.dart
```diff
@@ -58,6 +59,14 @@ abstract class PMFilter {
   ///  * [AssetPathEntity.lastModified].
   bool containsPathModified = false;
 
+  /// Whether to include hidden assets in the results.
+  ///
+  /// This option only takes effect on iOS.
```
AlexV525, line 64:
```suggestion
  /// This option only takes effect on iOS.
  /// Beginning with iOS 16, users can require authentication to view the
  /// hidden album, and the user setting is true by default. When true,
  /// the system doesn’t return hidden assets even if the option is true.
```

PR #1295, opened 2025-06-17
added islimited property to permission state extension

HubHive:
I would argue that having this property would and being able to do:

```
final permission = await PhotoManager.requestPermissionExtend();
if(permission.isLimted)
```

Rather than: 

`if(permission == PermissionState.limited)`


lib/src/internal/extensions.dart
```diff
@@ -13,7 +13,10 @@ extension PermissionStateExt on PermissionState {
 
   /// Returns `true` if the permission grants partial or full access to assets; otherwise, `false`.
   bool get hasAccess {
-    return this == PermissionState.authorized ||
-        this == PermissionState.limited;
+    return this == PermissionState.authorized || this == PermissionState.limited;
+  }
+
+  bool get isLimited {
```
AlexV525, line 19: I'm not sure about the `is` naming. @CaiJingLong got any comment here?

Issue #1257, opened 2025-01-14
[Filter Resource By Type] How to use RequestType to filter inside a specific album?


delfme:
### Version

3.5.2

### Platforms

dart

### Device Model

iphone 13 and android xiaomi mi lite 5g

### flutter info

```sh
Not relevant.
```

### How to reproduce?

I'm not sure if this qualifies as a bug, but I couldn't find an option to specify the media type (e.g., video or image) when fetching assets from an album using RequestType.

For example, I am currently using this code:
```
// currentAlbum is the 'Recents' folder in iOS or any other album
final assetEntities = await _currentAlbum.getAssetListPaged(page: pageIndex, size: pageSize);
for (int index = 0; index < assetEntities.length; index++) {
  if (assetEntities.elementAt(index).type == AssetType.video) {
    // do something with the video assets
  }
}
```
However, this approach is inefficient because it returns both images and videos. I don't see an option to specify an `AssetType` when calling `getAssetListPaged()`, as it only accepts `page` and `size` parameters.

Would you kindly explain if there's a way to filter by `AssetType` directly when fetching the assets from an album?

fluttercandies-dev:
**AI Summary:** The user wants to filter assets by type (e.g., video or image) within a specific album using `getAssetListPaged()`, but it lacks this functionality.

delfme:
I also tried this:
```
    final option = FilterOption();
    FilterOptionGroup filter = new FilterOptionGroup();
    filter.setOption(AssetType.video, option);
    currentAlbum = _currentAlbum.copyWith(filterOption: filter);
    final assetEntities = await _currentAlbum.getAssetListPaged(page: pageIndex, size: pageSize);
```
But I still get both images and videos from `getAssetListPaged`.

delfme:
This also is not working.
```
    currentAlbum = _currentAlbum.copyWith(type: RequestType.video);
    final assetEntities = await _currentAlbum.getAssetListPaged(page: pageIndex, size: pageSize);
```

delfme:
While doing these tests, I might also discovered a bug. If `type` is set to `RequestType.common`, then the sorting will be wrong (createAt sorting is no longer respected)
```
currentAlbum = _currentAlbum.copyWith(type: RequestType.common);
final assetEntities = await _currentAlbum.getAssetListPaged(page: pageIndex, size: pageSize);
```

I have no sorting issue if I simply:
```
[...]

PR #1305, opened 2025-08-14
feat: Fetch assets by specific type within an album

iosephmagno:
**Background & Need**
- The GitHub issue ”[Filter Resource By Type] How to use `RequestType` to filter inside a specific album? #1257” confirms the lack of this feature.
- Currently, developers can specify a `RequestType` globally, but not when fetching assets from a specific album.

**Proposed Enhancement:**
- Add an optional `RequestType`? `requestType` parameter to album-scoped methods like `getAssetListPaged`.
- If `requestType` is provided, pass it to native code; otherwise, fallback to `RequestType.common`.

**[entity.dart]**
- Allow overriding album type per call, enabling filtering for a specific asset type within an album.

**[PMManager.m]**
- Enforce type filtering. A change to native code is required only for the iOS platform.

**[main.dart]**
- Add `type` parameter to `getAssetListPaged()` for testing the new feature.

**[pubspec.yaml]**
- Bump build number.

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -125,6 +125,7 @@ class _SimpleExamplePageState extends State<_SimpleExamplePage> {
     final List<AssetEntity> entities = await _path!.getAssetListPaged(
       page: 0,
       size: _sizePerPage,
+      type: RequestType.common, // Switch to .video or .image to filter album assets by format
     );
     if (!mounted) {
       return;
@@ -140,6 +141,7 @@ class _SimpleExamplePageState extends State<_SimpleExamplePage> {
     final List<AssetEntity> entities = await _path!.getAssetListPaged(
       page: _page + 1,
       size: _sizePerPage,
+      type: RequestType.common,  // Switch to .video or .image to filter album assets by format
     );
     if (!mounted) {
       return;
--- a/ios/Classes/core/PMManager.m
+++ b/ios/Classes/core/PMManager.m
@@ -259,37 +259,71 @@
 
 - (NSArray<PMAssetEntity *> *)getAssetListPaged:(NSString *)id type:(int)type page:(NSUInteger)page size:(NSUInteger)size filterOption:(NSObject<PMBaseFilter> *)filterOption {
     NSMutableArray<PMAssetEntity *> *result = [NSMutableArray new];
-    
+
     PHFetchOptions *options = [PHFetchOptions new];
-    
+
     PHFetchResult<PHAssetCollection *> *fetchResult =
     [PHAssetCollection fetchAssetCollectionsWithLocalIdentifiers:@[id]
                                                          options:options];
     if (fetchResult && fetchResult.count == 0) {
         return result;
     }
-    
+
     PHAssetCollection *collection = fetchResult.firstObject;
     PHFetchOptions *assetOptions = [self getAssetOptions:type filterOption:filterOption];
+
+    // --- PATCH: enforce type filtering explicitly ---
+    // 'type' here is a bitmask based on PhotoManager's RequestType enum:
+    //   1 = image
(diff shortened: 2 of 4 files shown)


iosephmagno:
A practical example of this in Presence Messenger: the user can tap the video icon on the right side of the MediaPicker to filter videos in the currently selected album.

IMG_4133

lib/src/types/entity.dart
```diff
@@ -195,14 +195,15 @@ class AssetPathEntity {
...
+    RequestType? type, // allow overriding the album type per call
   }) {
     assert(albumType == 1, 'Only album can request for assets.');
     assert(size > 0, 'Page size must be greater than 0.');
     return plugin.getAssetListPaged(
       id,
       page: page,
       size: size,
-      type: type,
+      type: type ?? RequestType.common, // use album type by default
```
AlexV525, line 206: Shouldn't this be `type ?? this.type`?

pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: photo_manager
 description: A Flutter plugin that provides album assets abstraction management APIs on Android, iOS, macOS, and OpenHarmony.
 repository: https://github.com/fluttercandies/flutter_photo_manager
-version: 3.7.1
+version: 3.7.2
```
AlexV525, line 4: Please revert

ios/Classes/core/PMManager.m
```diff
@@ -259,37 +259,71 @@ - (void)injectAssetPathIntoArray:(NSMutableArray<PMAssetPathEntity *> *)array
...
+
     PHAssetCollection *collection = fetchResult.firstObject;
     PHFetchOptions *assetOptions = [self getAssetOptions:type filterOption:filterOption];
+
+    // --- PATCH: enforce type filtering explicitly ---
+    // 'type' here is a bitmask based on PhotoManager's RequestType enum:
+    //   1 = image
+    //   2 = video
+    //   3 = common  (image + video)  <-- THIS IS THE "COMMON" CASE
+    //   4 = audio
```
AlexV525, line 280: IIRC we have utils to convert these numbers into specific native enums. Is it true? The code here seems to be a bit of hard-coded also.

example/lib/main.dart
```diff
@@ -125,6 +125,7 @@ class _SimpleExamplePageState extends State<_SimpleExamplePage> {
     final List<AssetEntity> entities = await _path!.getAssetListPaged(
       page: 0,
       size: _sizePerPage,
+      type: RequestType.common, // Switch to .video or .image to filter album assets by format
```
AlexV525, line 128: If we imply the type from the path, then these are unnecessary.

lib/src/types/entity.dart
iosephmagno: It should, I don't remember why I forced a fallback.

ios/Classes/core/PMManager.m
iosephmagno: Oh, I didnt know about Utils. We should definitely use them.

Would you be so kind as to make these minor changes? I won't be able to work on this for a while. 

I quickly added this feat to our fork and is the only difference compared to the official repo.

AlexV525 (changes requested):
Upon reviewing the code, I'm unsure whether we need to update the Darwin implementation. The ability to filter corresponding assets appears to have been implemented; the only thing we need to add is parameters on the Dart side.

AlexV525:
I was not able to push to your PR branch, so please make modifications and confirm if they are working properly.

iosephmagno:
Aside from using a hardcoded type (I was not aware of utils to convert numbers into native enums), the code works correctly.

PR #1320, opened 2025-11-05
feat: Add relativePathAsync getter to AssetPathEntity

toyaji:
Add ability to get the relative path of an album/folder through AssetPathEntity.

Changes:
- Add relativePathAsync getter to AssetPathEntity (Dart layer)
- Implement platform channel method getPathRelativePath
- Add Android native implementation:
  - Android 10+ (API 29+): Uses MediaStore RELATIVE_PATH column
  - Android 9- (API 28-): Uses File.parent from DATA column
- iOS/macOS returns null (albums are logical, not physical folders)
- Add test UI in example app to display relative paths

Platform-specific behavior:
- Android: Returns relative path for device folders (e.g., 'DCIM/Camera')
- iOS: Returns null (PHAssetCollection logical albums have no file path)
- Special albums (All, Recent): Returns null on all platforms

### DIFF
--- a/android/src/main/kotlin/com/fluttercandies/photo_manager/core/PhotoManager.kt
+++ b/android/src/main/kotlin/com/fluttercandies/photo_manager/core/PhotoManager.kt
@@ -251,7 +251,7 @@ class PhotoManager(private val context: Context) {
     }
 
     fun getPathRelativePath(galleryId: String): String? {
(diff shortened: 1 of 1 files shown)


AlexV525 (commented):
Thanks for the improvement. A new CHANGELOG entry is also required before merging

Issue #1070, opened 2024-01-09
[Feature request] Support edit metadata on iOS with AAE sidecar support.


suprcharg3d:
### Platforms

iOS

### Description

Support getting both edited and original version, with edit history using AAE sidecar on iOS, instead of just the edited one with the edits baked in.

### Why

It would be really useful to be able to get both the original and edited files with edit history as to preserve the original files for future editing.

Skywalker8510:
bump.

Alex9779:
Is this still on some list? The label "missing requirements" is irritating me...

What is unclear?

AlexV525:
We might looking into it when we have enough bandwidth to.

Naworm:
Looking forward for this feature, here the discussion about it and useful ressources and maybe the missing requirements?
https://github.com/immich-app/immich/discussions/4750#discussioncomment-10128249

Skywalker8510:
its been a while since anything has been said about the status of this and I'm curious to know if there have been any new developments. not to rush development I'm just curious if anything has happened with this.

PR #1339, opened 2025-11-20
Feat: (iOS and macOS) Add interface to get hasAdjustments property and base adjustment files

john-trieu-nguyen:
**Summary**
Apple has the hasAdjustments property to indicate whether the asset contains adjustment data.
Base adjustment files are unaltered versions of the asset before any edits were applied. 

**Partially solves**
https://github.com/fluttercandies/flutter_photo_manager/issues/1070
Now able to get the 'original version' of files. Getting AAE sidecar would be a similar implementation but have left out for now.

**Motivation**
https://github.com/immich-app/immich/discussions/4750
https://github.com/immich-app/immich/issues/5818
Wanted ability to also upload the unedited image.

**Things to note**
PHAsset.hasAdjustments is only available on iOS and macOS 15.0+. Because this library targets 14.0, there is fallback implementation. The fallback implementation is very inefficient and could cause performance issues.

**References**
[PHAssetResourceType](https://developer.apple.com/documentation/photos/phassetresourcetype)
[PHAsset.hasAdjustments](https://developer.apple.com/documentation/photos/phasset/hasadjustments)

### DIFF
--- a/lib/src/internal/plugin.dart
+++ b/lib/src/internal/plugin.dart
@@ -325,6 +325,39 @@ class PhotoManagerPlugin with BasePlugin, IosPlugin, AndroidPlugin, OhosPlugin {
     return _channel.invokeMethod(PMConstants.mGetFullFile, params);
   }
 
+  Future<String?> getAdjustmentBaseFile(
+    String id, {
+    PMProgressHandler? progressHandler,
+    int subtype = 0,
+    PMDarwinAVFileType? darwinFileType,
+    PMCancelToken? cancelToken,
+  }) async {
+    final params = <String, dynamic>{
+      'id': id,
+      'subtype': subtype,
+      'darwinFileType': darwinFileType?.value ?? 0,
+    };
+    _injectProgressHandlerParams(params, progressHandler);
+    _setCancelToken(params, cancelToken);
+    return _channel.invokeMethod(PMConstants.mGetAdjustmentBaseFile, params);
+  }
+
+    Future<String?> getAdjustmentBaseLivePhotoFile(
+    String id, {
+    PMProgressHandler? progressHandler,
+    int subtype = 0,
+    PMDarwinAVFileType? darwinFileType,
+    PMCancelToken? cancelToken,
+  }) async {
+    final params = <String, dynamic>{
+      'id': id,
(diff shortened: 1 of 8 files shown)


lib/src/internal/plugin.dart
AlexV525, line 1: Can we put those new methods under `IosPlugin`?

lib/src/types/entity.dart
```diff
@@ -860,6 +851,68 @@ class AssetEntity {
...
+  ///  * TypeAdjustmentBaseVideo || TypeVideo
+  Future<File?> getAdjustmentBaseFile({
+    PMProgressHandler? progressHandler,
+    int subtype = 0,
+    PMDarwinAVFileType? darwinFileType,
+    PMCancelToken? cancelToken,
+  }) async {
+
+    if (Platform.operatingSystem != 'ios' &&
+        Platform.operatingSystem != 'macos') {
```
AlexV525, line 868: Use `Platform.isIOS` and `Platform.isMacOS`

lib/src/types/entity.dart
```diff
@@ -860,6 +851,68 @@ class AssetEntity {
...
+  ///  * iOS/macOS: `PHAssetResource` with type:
+  ///  * TypeAdjustmentBasePhoto || TypePhoto
+  ///  * TypeAdjustmentBaseVideo || TypeVideo
+  Future<File?> getAdjustmentBaseFile({
+    PMProgressHandler? progressHandler,
+    int subtype = 0,
+    PMDarwinAVFileType? darwinFileType,
+    PMCancelToken? cancelToken,
+  }) async {
+
```
AlexV525, line 866: The file doesn't seem to be well-formatted.

AlexV525 (commented):
Sorry for the late review.

AlexV525:
Hi @john-trieu-nguyen are you still working on this?

AlexV525:
Superseded by #1407, which implements this behind the new `AssetEntity.darwin` / `AssetPathEntity.darwin` namespaces (grouping Darwin-only PhotoKit reads instead of adding members onto the shared entity classes). Closing in favor of that PR.

PR #1352, opened 2025-12-04
Refactor getSortOrder in IDBUtils.kt to ensure valid SQL query construction by providing a default ORDER BY clause when no filter option is specified.

esentis:
Tested with Xiaomi Mi Mix 2

### Fix Android album query SQL ( DBUtils#getAssetPathList )

* Previously, the selection string was concatenated without a separating space/parenthesis, generating invalid SQL like `BUCKET_ID IS NOT NULLbucket_id IS NOT NULL … GROUP BY`. On stricter OEM MediaStore implementations this caused SQLiteException: no such column: NULLbucket_id when loading albums.
* Updated to build the selection as `BUCKET_ID IS NOT NULL` plus the filter clause, with proper spacing and grouping, eliminating the malformed column token.

### Ensure valid sort clause for paging ( IDBUtils#getSortOrder )

* When no explicit order was provided, the plugin emitted `ORDER BY LIMIT … OFFSET …`, which is invalid SQL and produced `near "LIMIT": syntax error` on some devices when paging assets.
* Added a default order (`DATE_ADDED DESC, DATE_MODIFIED DESC`) before `LIMIT/OFFSET`, so paging queries are syntactically correct across devices.

### DIFF
--- a/android/src/main/kotlin/com/fluttercandies/photo_manager/core/utils/DBUtils.kt
+++ b/android/src/main/kotlin/com/fluttercandies/photo_manager/core/utils/DBUtils.kt
@@ -38,7 +38,7 @@ object DBUtils : IDBUtils {
 
         var selection: String = "${MediaStore.MediaColumns.BUCKET_ID} IS NOT NULL"
         if (where != null) {
-            selection += "$selection $where"
+            selection += " $where"
         }
         selection += ") GROUP BY (${MediaStore.MediaColumns.BUCKET_ID}"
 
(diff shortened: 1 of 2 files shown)


AlexV525 (commented):
@esentis Thanks for the catch. Could you add corresponding tests to ensure it works as desired? It would be helpful if you could make screen records to indicate how the previous implementation is not working and how the fix works. Additionally, consider adding a page as a regression tracker under https://github.com/fluttercandies/flutter_photo_manager/tree/main/example/lib/page/developer/issues_page

esentis:
Ok it seems that the changes made with `3.8.3` fixed my issue so I'm closing this.

PR #1403, opened 2026-06-08
Migrate photo_manager to opt built-in Kotlin

ubxty:
Migrates the `photo_manager` subpackage to use built-in Kotlin per the official Flutter migration guide:
https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors

This PR scopes the changes to the `photo_manager` package only (the repository is a monorepo that previously hosted multiple subpackages; the `photo_manager` source tree is now at the repository root).

AGP 9.0 will remove support for plugins that apply the Kotlin Gradle Plugin. This PR removes the legacy `apply plugin: "kotlin-android"` line and the KGP classpath from `android/build.gradle`, and the same from the example app's `example/android/app/build.gradle`. The `kotlinOptions { jvmTarget = ... }` block is replaced with the modern `kotlin { compilerOptions { jvmTarget = ... } }` DSL.

The example app builds cleanly with `flutter build apk --debug` on Flutter 3.44.1 with no 'plugin applies Kotlin Gradle Plugin' warning.

Bumps the minimum Flutter version to 3.44.0 and Dart SDK to 3.12.0, as required by the new `kotlin.compilerOptions` DSL, and the package version to 3.10.0.

Flips `android.builtInKotlin=true` and `android.newDsl=true` in `example/android/gradle.properties` so the example app picks up the built-in Kotlin compiler.

Also fixes a pre-existing typo in the plugin's `android/build.gradle` where the declared Android `namespace` (`com.flutterandies.photo_manager`) did not match the actual Kotlin source package (`com.fluttercandies.photo_manager`). The mismatch was masked by the legacy KGP reading the package from the source directories; built-in Kotlin enforces the declared `namespace`, so the typo is corrected to match the real code and the auto-generated `GeneratedPluginRegistrant.java`.

Fixes the warning: "Your app uses the following plugins that apply Kotlin Gradle Plugin (KGP)".


AlexV525:
Could you do this another way that doesn't require bumping the minimum Flutter version?

PR #1424, opened 2026-07-16
✨ Add Restore From Trash on Android

### LINKED ISSUE #1423: [Feature request] Add `restoreFromTrash` for Android trashed items
### Platforms

Android

### Description

Support already exists for sending files to trash, restore is straightforward.

### Why

This is one of the few APIs we require to add support for system trash.

aspizu:
Closes #1423

### DIFF
--- a/lib/src/internal/plugin.dart
+++ b/lib/src/internal/plugin.dart
@@ -1075,6 +1075,20 @@ mixin IosPlugin on BasePlugin {
 }
 
 mixin AndroidPlugin on BasePlugin {
+  Future<List<String>> restoreFromTrash(List<AssetEntity> list) async {
+    assert(
+      Platform.isAndroid || Platform.environment.containsKey('FLUTTER_TEST'),
+    );
+    final List result = await _channel.invokeMethod(
+      PMConstants.mRestoreFromTrash,
+      <String, dynamic>{
+        'ids': list.map((e) => e.id).toList(),
+        'types': list.map((e) => e.typeInt).toList(),
+      },
+    );
+    return result.cast<String>();
+  }
+
   Future<void> forceOldApi() async {
     assert(Platform.isAndroid);
     if (Platform.isAndroid) {
--- a/lib/src/internal/plugin.dart
+++ b/lib/src/internal/plugin.dart
@@ -1076,9 +1076,7 @@ mixin IosPlugin on BasePlugin {
 
 mixin AndroidPlugin on BasePlugin {
   Future<List<String>> restoreFromTrash(List<AssetEntity> list) async {
-    assert(
-      Platform.isAndroid || Platform.environment.containsKey('FLUTTER_TEST'),
-    );
+    assert(Platform.isAndroid);
     final List result = await _channel.invokeMethod(
       PMConstants.mRestoreFromTrash,
       <String, dynamic>{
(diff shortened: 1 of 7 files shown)


AlexV525:
Thanks for the contribution. I was wondering if we could make `AndroidAsset` so that the method could be accessed by `entity.android.restoreFromTrash`. WDYT? We can implement this after the PR if you'd like to.

aspizu:
we would prefer to continue using the current `editor.android.moveToTrash`, `editor.android.restoreFromTrash` APIs. We don't have any opinions on the shape of a newer API.

AlexV525:
## Automated review

**Correctness — looks solid**
- Correctly passes `false` to `createTrashRequest` (vs `true` in `moveToTrash`), reusing the existing `androidRDeleteRequestCode` / `handleAndroidRDelete` activity-result plumbing.
- Builds the asset URI from client-supplied `id` + `type` (`MediaStoreUtils.getUri`) instead of `photoManager.getUri(id)`, which is the right call — `getUri(id)` queries the DB and excludes trashed items, so it would throw for a trashed asset. Good catch, and it's called out in the Dart doc comment.
- `AssetType` index mapping (`other=0, image=1, video=2, audio=3`) lines up with `MediaStoreUtils.convertTypeToMediaType`, and the `ids.size == types.size` guard is a sensible addition.
- Matches the sibling `moveToTrash`/`deleteWithIds` handlers in API-level gating and error handling, and even fixes a pre-existing copy/paste bug where `moveToTrash`'s catch logged "deleteWithIds failed".

**Pre-existing design issues surfaced (not regressions, just now with a third consumer)**
- `restoreFromTrashInApi30` shares `androidRDeleteRequestCode`/`androidRHandler` with `deleteInApi30` and `moveToTrashInApi30`. Overlapping calls (caller not awaiting) would clobber the handler and leave the first Future unresolved.
- If `activity` is null, `startIntentSenderForResult` is a no-op and the Dart Future hangs forever — same latent issue as the other two trash/delete paths.

**Gaps worth addressing before merge**
1. Missing `CHANGELOG.md` entry (Unreleased → Features), which every other feature PR in this repo includes.
2. Missing README documentation — there's a "Move entities to trash" section under "Features for Android only" with a code sample, but no matching "Restore entities from trash" section for this new public API.
3. In `plugin.dart`, the new method uses `assert(Platform.isAndroid || Platform.environment.containsKey('FLUTTER_TEST'))`, while every other platform-gated method in the file uses a plain `assert(Platform.isAndroid)` and gates tests via `skip: !...` instead. Worth aligning with the existing convention rather than adding a `FLUTTER_TEST` bypass.
4. Minor: `restoreFromTrash` was placed in the `AndroidPlugin` mixin with a runtime assert, while `moveToTrash` lives in the platform-agnostic `BasePlugin` mixin with no assert — a small inconsistency between two very similar operations.

Overall the core implementation is correct and shows real understanding of the MediaStore trash API's query limitations. Main asks are the CHANGELOG/README docs and the assert-style nit.

aspizu:
could u add the changelog/docs entries? is there anything specific you want me to fix before we merge?

lib/src/internal/plugin.dart
```diff
@@ -1075,6 +1075,20 @@ mixin IosPlugin on BasePlugin {
 }
 
 mixin AndroidPlugin on BasePlugin {
+  Future<List<String>> restoreFromTrash(List<AssetEntity> list) async {
+    assert(
+      Platform.isAndroid || Platform.environment.containsKey('FLUTTER_TEST'),
+    );
```
AlexV525: Please keep the behavior here. Check `Platform.isAndroid` only, and remove the corresponding test. The added test is not necessary.

PR #1425, opened 2026-07-16
✨ Add isTrashed field and query trashed items when includeTrashed=true in android

### LINKED ISSUE #1426: [Bug report] AssetEntity.fromId() fails for trashed files
### Version

latest

### Platforms

Android

### Device Model

N/A

### flutter info

```sh
N/A
```

### How to reproduce?

call sendToTrash on an AssetEntity
then call AssetEntity.fromId(id) again which will fail

### Logs

```sh

```

### Example code (optional)

```Dart

```

### Contact

_No response_

aspizu:
Closes #1426

### DIFF
--- a/android/src/main/kotlin/com/fluttercandies/photo_manager/core/utils/AndroidQDBUtils.kt
+++ b/android/src/main/kotlin/com/fluttercandies/photo_manager/core/utils/AndroidQDBUtils.kt
@@ -238,7 +238,8 @@ object AndroidQDBUtils : IDBUtils {
             keys(),
             selection,
             args,
-            null
+            null,
+            includeTrashed = true
         )
         cursor.use {
             return if (it.moveToNext()) it.toAssetEntity(context, checkIfExists)
--- a/android/src/main/kotlin/com/fluttercandies/photo_manager/core/utils/IDBUtils.kt
+++ b/android/src/main/kotlin/com/fluttercandies/photo_manager/core/utils/IDBUtils.kt
@@ -9,6 +9,7 @@ import android.database.MatrixCursor
 import android.media.MediaMetadataRetriever
 import android.net.Uri
 import android.os.Build
+import android.os.Bundle
 import android.os.Environment
 import android.provider.MediaStore
 import android.provider.MediaStore.MediaColumns.BUCKET_DISPLAY_NAME
@@ -21,6 +22,7 @@ import android.provider.MediaStore.MediaColumns.DISPLAY_NAME
 import android.provider.MediaStore.MediaColumns.DURATION
 import android.provider.MediaStore.MediaColumns.HEIGHT
 import android.provider.MediaStore.MediaColumns.IS_FAVORITE
+import android.provider.MediaStore.MediaColumns.IS_TRASHED
 import android.provider.MediaStore.MediaColumns.MIME_TYPE
 import android.provider.MediaStore.MediaColumns.ORIENTATION
 import android.provider.MediaStore.MediaColumns.RELATIVE_PATH
@@ -66,6 +68,7 @@ interface IDBUtils {
         ).apply {
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) add(DATE_TAKEN) // 拍摄时间
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) add(IS_FAVORITE)
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) add(IS_TRASHED)
         }
 
         val storeVideoKeys = mutableListOf(
@@ -85,6 +88,7 @@ interface IDBUtils {
         ).apply {
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) add(DATE_TAKEN) // 拍摄时间
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) add(IS_FAVORITE)
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) add(IS_TRASHED)
         }
 
         val typeKeys = arrayOf(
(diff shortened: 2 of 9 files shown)


aspizu:
im not exactly sure about the field `isTrashed` and the behaviour change of getAssetEntity

AlexV525:
## Automated review

**Overview**
Fixes a real bug: `AssetEntity.fromId(id)` fails on Android when the asset has been moved to the system trash, because `ContentResolver.query()` excludes trashed items by default on Android 11+ (API 30). Adds `isTrashed` to `AssetEntity` (Dart + Kotlin), threads it through `ConvertUtils`, adds `IS_TRASHED` to the R+ projection/cursor read (mirroring `isFavorite`), and adds an `includeTrashed` param to `IDBUtils.logQuery` that builds a `Bundle` with `MediaStore.QUERY_ARG_MATCH_TRASHED = MATCH_INCLUDE` when true.

**Correctness — looks solid**
- Scoped correctly: only the "get by id" path (`AndroidQDBUtils.getAssetEntity`) opts into `MATCH_INCLUDE`; general listing/album queries are untouched, so trashed items won't leak into normal listings.
- Android version gating (`Build.VERSION_CODES.R`) is applied consistently wherever `IS_TRASHED` is touched.
- The `Bundle`-based query construction matches Google's documented pattern for including trashed rows via the 4-arg `query(uri, projection, queryArgs, cancellationSignal)` overload.
- `equals`/`hashCode` correctly updated to include `isTrashed`.

**Worth flagging**
- `includeTrashed` is hardcoded `true` unconditionally in `AndroidQDBUtils.getAssetEntity`, so *every* by-id lookup (not just `fromId`) now silently matches trashed rows — used by several other call paths (favorite/delete/move flows, post-save fetch). Likely intentional/desired, but no caller can opt out, and it'd help to call this out explicitly (comment or PR description) for future readers.
- `AssetEntity`'s new `isTrashed` param was inserted mid-constructor (after `isFavorite`) rather than appended at the end. The one real call site was updated correctly, but appending new fields at the end is safer for a `data class` with many positional/optional params.
- `convert_utils.dart` hardcodes the raw string `'is_trashed'`, whereas the neighboring `isFavorite` goes through the `CustomColumns.base.isFavorite` abstraction. Functionally fine since this is Android-only, but diverges from the file's established convention.

**Test coverage**
- New tests only cover the `ConvertUtils` mapping round-trip (Kotlin + Dart). The actual new logic — the `Bundle`/`QUERY_ARG_MATCH_TRASHED` branch in `IDBUtils.logQuery` — has no test coverage. Understandably hard without Robolectric/instrumentation, but worth flagging since it's the one genuinely new/risky code path.

**Gaps worth addressing before merge**
1. Missing `CHANGELOG.md` entry under Unreleased/Fixes referencing the `AssetEntity.fromId()` trashed-asset fix.

Overall: small, targeted fix that correctly resolves the linked bug without touching general listing behavior. Main asks are the CHANGELOG entry and (ideally) a note on why `includeTrashed` is unconditional in `getAssetEntity`.

android/src/main/kotlin/com/fluttercandies/photo_manager/core/utils/IDBUtils.kt
```diff
@@ -786,7 +791,17 @@ interface IDBUtils {
...
 
         try {
-            val cursor = query(uri, projection, selection, selectionArgs, sortOrder)
+            val cursor = if (includeTrashed && Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+                val queryArgs = Bundle().apply {
+                    putString(ContentResolver.QUERY_ARG_SQL_SELECTION, selection)
+                    putStringArray(ContentResolver.QUERY_ARG_SQL_SELECTION_ARGS, selectionArgs)
+                    putString(ContentResolver.QUERY_ARG_SQL_SORT_ORDER, sortOrder)
+                    putInt(MediaStore.QUERY_ARG_MATCH_TRASHED, MediaStore.MATCH_INCLUDE)
+                }
```
AlexV525, line 800: Is `Bundle` required to be used here?

lib/src/utils/convert_utils.dart
```diff
@@ -122,6 +122,7 @@ class ConvertUtils {
       duration: data['duration'] as int? ?? 0,
       orientation: data['orientation'] as int? ?? 0,
       isFavorite: data[CustomColumns.base.isFavorite] as bool? ?? false,
+      isTrashed: data['is_trashed'] as bool? ?? false,
```
AlexV525: Could `is_trashed` be a field for custom columns?

android/src/main/kotlin/com/fluttercandies/photo_manager/core/utils/AndroidQDBUtils.kt
```diff
@@ -238,7 +238,8 @@ object AndroidQDBUtils : IDBUtils {
             keys(),
             selection,
             args,
-            null
+            null,
+            includeTrashed = true
```
aspizu, line 242: **MEDIUM** — This only enables trash matching in `AndroidQDBUtils`, but the public `PhotoManager.forceOldApi()` switch makes Android 11+ use `DBUtils`. Its `getAssetEntity()` still calls `logQuery` without `includeTrashed`, so MediaStore filters the trashed row and `AssetEntity.fromId()` still returns `null` in the fallback mode meant for devices where the new API does not work. Pass `includeTrashed = true` from `DBUtils.getAssetEntity()` as well—the helper already gates it to API 30+—and add a test for the forced-old-API route.

aspizu (commented):
Automated code review: high- and medium-severity findings only.

android/src/main/kotlin/com/fluttercandies/photo_manager/core/utils/IDBUtils.kt
aspizu: i think yes

android/src/main/kotlin/com/fluttercandies/photo_manager/core/utils/AndroidQDBUtils.kt
aspizu: not sure if i should fix this

AlexV525: Not applicable.

AlexV525 (approved):

PR #1430, opened 2026-08-02
🐛 Fix App Store rejection from PHAsset.filename KVC access

### LINKED ISSUE #1429: [Bug report] App Store rejection from PHAsset filename access
### Version

3.9.0

### Platforms

iOS, macOS

### Description of the issue

The App Store review process may rejects apps compiled against this plugin as, according to their scanning, there is code in the app that can access `PHAsset.filename`, which is either private or deprecated. 

I believe the culprit could be in the implementation for `title` in [PHAsset+PM_COMMON.m](https://github.com/fluttercandies/flutter_photo_manager/blob/7add38ce3db765679a4b52ddf0b39efa23344b1e/darwin/photo_manager/Sources/photo_manager/core/PHAsset+PM_COMMON.m), in particular this line:

```objective-c
NSString *result = [self valueForKey:@"filename"];
```

which tries to access the `filename` property / selector on `PHAsset`.

It seems that this code was touched in 8126cb51 a few years ago, and that before this used to access `originalFilename` instead. `PHAsset.originalFilename` may also be private or deprecated, as I cannot see any references to it in Apple's `PHAsset` header.

Currently, this selector resorts to accessing `PHAssetResource.originalFilename` if `PHAsset.filename` returns an exception. Would it make sense to always reach directly for the former and remove the attempt to surreptitiously call the latter? I'm happy to make this change if you think the approach is sound.

### Contact

https://github.com/davila-vilanova

CaiJingLong:
## Summary

- The Darwin `title` method reached the non-public `PHAsset.filename` selector via KVC (`[self valueForKey:@"filename"]`), which App Store's static private-API scanner flags, causing app rejections (#1429).
- Drop the `@try`/KVC attempt and go straight to the public `PHAssetResource.originalFilename` API — the same lookup the `@catch` branch already used as a fallback, and the same API the sibling `filenameWithOptions:isOrigin:fileType:` uses exclusively.
- Update `CHANGELOG.md` under `## Unreleased`.

Closes #1429

## Problem

`[self valueForKey:@"filename"]` is KVC: it invokes the `-filename` selector on a `PHAsset`. That selector is **not** part of the public Photos API (absent from `PHAsset.h` in both the iPhoneOS and MacOSX SDK headers), so it is private/undocumented. App Store's static scan detects the reference and rejects affected apps. The pre-existing `@catch` fallback already used the public `PHAssetResource.originalFilename`, so the KVC attempt was both risky and redundant.

## Change

`darwin/photo_manager/Sources/photo_manager/core/PHAsset+PM_COMMON.m` — `title`:
- Removed the `@try { [self valueForKey:@"filename"] } @catch { ... }` block.
- `title` now iterates `PHAssetResource.assetResourcesForAsset:` and returns `originalFilename` for the matching photo/video resource, falling back to the first resource, then `@""` — the exact logic that lived in the `@catch` branch.
- Added a comment documenting why KVC into `PHAsset.filename` is avoided.

No public Dart API, signature, or cross-platform semantics change. The returned value is the same `PHAssetResource.originalFilename` the fallback already produced.

## Verification

- `clang -fsyntax-only -fobjc-arc -isysroot "$(xcrun --sdk macosx --show-sdk-path)" -fmodules` on the changed file: compiles clean (only pre-existing, unrelated `UTTypeCopyPreferredTagWithClass`/`kUTTagClassMIMEType` deprecation warnings in `mimeType`).
- `dart format . -o none --set-exit-if-changed`: 0 changed.
- `flutter analyze lib`: No issues found.
- Confirmed no remaining `valueForKey:@"filename"` / `valueForKey:@"originalFilename"` references under `darwin/`.

## Risk

Low. Change is confined to the Darwin `title` accessor; behavior matches the previous `@catch` fallback path that already ran whenever the private selector threw. No new dependencies, abstractions, or files.

## Notes

- Could not reproduce an actual App Store rejection (requires submitting for review); the underlying claim — a reference to a private `PHAsset` selector via KVC — is verified and removed.
- Follows `AGENTS.md` (gitmoji commit, minimal diff, CHANGELOG under `## Unreleased`, no version-floor change).

> This comment was generated by AI agent omp (model: devin/glm-5-2).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,9 @@ To know more about breaking changes, see the [Migration Guide][].
 
 ## Unreleased
 
-*None.*
+**Fixes**
+
+- Avoid App Store rejections caused by referencing the private `PHAsset.filename` selector. The Darwin `title` lookup now uses the public `PHAssetResource.originalFilename` API directly instead of KVC'ing into the non-public `filename` selector.
 
 ## 3.11.0
 
--- a/darwin/photo_manager/Sources/photo_manager/core/PHAsset+PM_COMMON.m
+++ b/darwin/photo_manager/Sources/photo_manager/core/PHAsset+PM_COMMON.m
@@ -41,30 +41,26 @@
 }
 
 - (NSString *)title {
+    // Avoid KVC into the private `PHAsset.filename` selector; App Store's static
+    // scanner flags references to it. `PHAssetResource.originalFilename` is the
+    // only public API for obtaining an asset's filename.
     PMLogUtils *logger = [PMLogUtils sharedInstance];
-    [logger info:@"get title start"];
-    @try {
-        NSString *result = [self valueForKey:@"filename"];
-        [logger info:@"get title from kvo"];
-        return result;
-    } @catch (NSException *exception) {
-        [logger info: @"get title from PHAssetResource"];
-        NSArray *array = [PHAssetResource assetResourcesForAsset:self];
-        for (PHAssetResource *resource in array) {
-            if ([self isImage] && resource.type == PHAssetResourceTypePhoto) {
-                return resource.originalFilename;
-            } else if ([self isVideo] && resource.type == PHAssetResourceTypeVideo) {
-                return resource.originalFilename;
-            }
+    [logger info:@"get title from PHAssetResource"];
+    NSArray *array = [PHAssetResource assetResourcesForAsset:self];
+    for (PHAssetResource *resource in array) {
+        if ([self isImage] && resource.type == PHAssetResourceTypePhoto) {
+            return resource.originalFilename;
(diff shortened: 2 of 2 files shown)


AlexV525 (approved):
Confirmed. The reporter provided the rejection screenshot (#1429) — App Store review explicitly named `PHAsset.filename` as the private API. My earlier doubt about whether KVC (`valueForKey:@filename`) evades the scanner was wrong: Apple's detection catches this path even though `strings`/`otool -ov` cannot find the selector post-hoc.

This change is correct and now justified by evidence. The fix removes the only confirmed-offending reference (`filename`) and routes to the public `PHAssetResource.originalFilename`, matching the existing `@catch` fallback and the sibling `filenameWithOptions:` method.

